PR #22, opened 2020-07-02
Document setup of env vars

jgadsden:
This pull request documents how to set up the environment variables ofr Threat Dragon webapp version. It has beenn the subject of a few issues and it would be good to give this step by step account of how to set these vars


jgadsden:
Can merge in after a couple of reviews?

PR #42, opened 2020-11-30
feat: added dockerfile

omerlh:
**Summary**

Really small contribution - added docker file to simplify the process of deploying threat dragon

Add dockerfile

**Other info**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM node:14 as build
+
+WORKDIR /app
+
+COPY package.json .
+
+RUN npm install
+
+FROM gcr.io/distroless/nodejs:14
+COPY --from=build /app /app
+WORKDIR /app
+COPY . . 
+CMD ["server.js"]


jgadsden (approved):
Hello @omerlh - thankyou, this is very good to see.

Do we want to add some instructions in our readme notes to how to use it? I am familiar with docker, and I know other users are not.

We may also want to provide some help with the environment variables that we need to run threat dragon ... probably the subject of another pull request?

omerlh:
I'm actually working on a helm chart for deployment, hope I could OSS it.

Maybe we can just auto deploy it to dockerhub?

jgadsden:
Hello @omerlh - we need to setup the environment variables for the container, do you have instructions for doing this for the docker container?

https://github.com/OWASP/threat-dragon#environment-variables

so far I have built the docker image using `docker build . --tag threatdragon` and can run it using port mapping `docker run --name threatdragon -p3000:3000 threatdragon:latest` but it complains about the environment variables.

omerlh:
We could in theory add `ENV` instruction - but I think this is more about documentation (run with `-e` to pass env var / add the relevant helm chart)

jgadsden:
Cool, understood - I will add some documentation. Thanks again for this PR

omerlh:
Thank you!

PR #46, opened 2020-12-03
chore: refactor server code to use typescript

omerlh:
**Summary**

Move all server code to be written using typescript, to make maintenance and future development easier.

chore: refactor to use typescript

**Other info**


jgadsden:
Hello @omerlh - many  thanks for the pull request, very interesting to see the changes needed to support typescript. 

@mike-goodwin and @andk123 could you review this PR? I have to admit that I know little about typescript, other than that it is a superset of javascript, and have avoided it because it is created and supported by Microsoft. My instinctive reaction is to keep to javascript so that typescript-only code does not creep in, but I realise my thinking is limited in this matter and so I think it is over to you two to review the PR?

omerlh:
While it is true that TypeScript was created by Microsoft, it has many advantages - especially, type safety, which makes maintaining projects a lot easier. TypeScript become the standard for many project, slowly replacing JavaScript, I am not sure we should avoid that...

andk123:
Very interesting! Did you use a tool to generate the typescript code @omerlh? I see the directory structure also changed a bit. I am not a professional in Typescript either, so I'm not sure if I can review the changes (although at a first glance, there is no functional change, but more a design change). Also, what is the idea behind keeping one part of the code in javascript and the other part in typescript? Just wondering

omerlh:
I did the migration manually, so I migrated most of the code, but not all of it - I think this can be another iteration...

jgadsden:
Closing this pull request  until we decide what to do about Typescript

PR #67, opened 2021-04-25
Feature/separate server from site

lreading:
**Summary**
Separate the front-end code from the back-end code

Created a separate directory at the root of the project for `td.site`, and moved `td` to `td.server`.  Updated all local file references including scripts in `package.json. `

**Other info**
I think it would make sense to move the front-end to its own directory.  This will allow us to migrate the back-end code to ES6 independently from any front-end code changes (eg Vue migration, e2e tests, etc). The front-end is using browserify and is slated to be migrated to Vue in the future.  Having the front-end code fully separated from the back-end allows more options when moving to ES6, using transpalation tools like babel.  It will also allow for more of a pragmatic approach to refactoring the back-end, where less work will need to be done in a single sweep.

I'm very open to feedback and discussion on this.


package.json
```diff
@@ -4,21 +4,21 @@
   "private": false,
   "scripts": {
     "build": "npm run-script bundle && npm run-script bundle-minify && npm run-script bundle-css && npm run-script bundle-minify-css && npm run-script copy-fonts",
-    "build-templates": "browserify -t [browserify-ng-html2js --module templates] ./td/public/app/templates.build.js -o ./td/public/app/templates.js",
-    "copy-fonts": "./node_modules/.bin/copyfiles -f ./node_modules/font-awesome/fonts/*.* ./node_modules/bootstrap/dist/fonts/*.* ./td/public/fonts",
-    "bundle-css": "./node_modules/.bin/rework-npm ./td/public/content/app.css -o ./td/public/content/threatdragon.css",
-    "minify-css": "./node_modules/.bin/uglifycss ./td/public/content/threatdragon.css > ./td/public/content/threatdragon.min.css",
+    "build-templates": "browserify -t [browserify-ng-html2js --module templates] ./td.site/app/templates.build.js -o ./td.site/app/templates.js",
```
lreading, line 7: Is this script needed anymore? I couldn't find `templates.js` or `templates.build.js`

jgadsden: agreed, this can be safely removed

jgadsden (approved):
This is very good to see, many thanks to @lreading . Tested it just now it works as before.

This is one of the first steps in the version 2.0 roadmap, which will make it easier to provide an API to Threat Dragon. This has been requested in some verbal conversation I have  had with users, which will help them use it in CI / CD pipelines and the like

jgadsden:
There are two errors from the code scanning. I do not think these were introduced in this pull request, they seem to have always been present. @lreading did you want to confirm this?

Looks good to me for merging

lreading:
It looks to be complaining about the front-end `threatdragon.js`: there weren't any substantial changes there (if any).  My guess is that where the file is in a different path it's identifying it as a *new* file and alerting on that?

jgadsden:
Agreed, we should go ahead and merge. We will be probably be revising all this code for version 2.0 so it may not be in the scan when that time comes.

Issue #80, opened 2021-04-28
bring docs and dev notes into threat dragon repo


jgadsden:
**Describe what problem your feature request solves**
At present much of the dev notes are in the core directory, because that is common to both web app and desktop app. 

**Describe the solution you'd like**
It would be good to bring docs and dev notes into one place, the threat dragon repo, because core repo may be combined with threat dragon repo soon. Also desktop app will soon be an electron wrapper for threat dragon repo

**Additional context**
It would be good to add docs repo https://github.com/threatdragon/threatdragon.github.io as a submodule to threat dragon repo

PR #83, opened 2021-04-28
Feature/75 front end package json

### LINKED ISSUE #75: Create independent package.json for front-end
Create a separate package.json for the back-end code

As discussed in https://github.com/OWASP/threat-dragon/discussions/69#discussioncomment-660564 and https://github.com/OWASP/threat-dragon/issues/70#issuecomment-827029562 we should have separate package.json files for the front-end and back-end as they have different needs.

**Additional context**
After a little initial investigation, this will be a non-trivial amount of work to attack all at once.  This issue is specific to the front-end code.
Scripts and tests will need to be migrated, and this is a great opportunity to make some changes to the build process as well.  Currently, we are keeping the minified / bundled code in source control.  Additionally, we package the unminified code alongside the minified code. We can remove the minified code from source control, and only package the minified code as part of the build process.

lreading:
**Summary**
Create an independent package.json for the front-end code

Closes #75 - Moved front-end code to a different folder structure and it now has its own package.json.  Updated build process and removed minified files from source control.

**Other info**
- Dockerfile builds front-end and works with the new directory structure
- Readme reflects changes made and removed some duplicated information (installation/build had the same info)
- Changed root package.json to just have convenience scripts for building and running
- Relative paths updated as necessary
- Front-end uses webpack now for bundling, and outputs into `dist` (ignored in gitignore). 
  - The package.json was becoming unruly when trying to update the build process using the existing tools, too many tasks
  - There is no dev server configured at this time (could be done in the future, but maybe not needed where there's a full rewrite planned)
  - The bundles that are created are not fully optimized, they could be chunked better.
- The root package.json now does the following:
  - The main scripts (`install`, `test`, `start`) were changed to be backwards compatible: install will install td.site and td.server npm dependencies via postinstall script. test will run tests for td.site and td.server. start will build td.site and then run td.server
- Some of the required resources from node_modules on the front-end are copied to the dist directory by webpack; local references were updated to reflect that


lreading:
I believe this requires a documentation update.  Do you want me to wait for #80 to be done before working on that?

jgadsden (approved):
Very good to see this, and great to see the README kept up to date

PR #107, opened 2021-05-03
Chore/server es6 refactor config

lreading:
**Summary**
Refactor the td.server/config to use ES6 syntax 

chore: Refactor td.server/config to use ES6 syntax

**Other info**
All config files are now using es6 syntax.  Improved test coverage a bit as well.


td.server/src/config/parsers.config.js
```diff
@@ -1,18 +1,12 @@
...
-var cookieParser = require('cookie-parser');
-var bodyParser = require('body-parser');
-
-var parsers = function (app) {
-   
-    app.use(bodyParser.json());
-    app.use(bodyParser.urlencoded({ extended: false }));
-    app.use(cookieParser());  
-    
+const config = (app) => {
```
lreading, line 4: body-parser has been depreciated.  The official recommendation is to use `express.json()` and `express.urlencoded` functions.  I removed the body-parser dependency as well.

td.server/test/config/parsers.config_spec.js
```diff
@@ -38,15 +40,15 @@ describe('request parser tests', () => {
         const body = {id: id, collection: collection, nested: nested};
         
         app.post('/', function(req, res) {
-            expect(req.body["collection[0]"]).to.eq(collection[0]);
-            expect(req.body["collection[1]"]).to.eq(collection[1]);
-            expect(req.body["collection[2]"]).to.eq(collection[2]);
-            expect(req.body["nested[n1]"]).to.eq(nested.n1);
-            expect(req.body["nested[n2]"]).to.eq(nested.n2);
+            expect(req.body.collection[0]).to.eq(collection[0]);
```
lreading, line 43: Removing the bodyParser dependency and using the built-in express parser changed how this works a bit.  Using dot-notation works just fine.

I searched the codebase and didn't find any instances of `req.body[`, so I believe this to be a non-breaking change.

jgadsden (approved):
I thought the best thing was to run this branch from the command line and from docker. Both work as before, so definitely approve.  Thanks as always @lreading

PR #180, opened 2021-05-16
V2/continuous deployment

lreading:
**Summary**
Hardening the CI/CD Pipelines

Browserstack only runs nightly, deploys on run on  the v2-deployment branch, smoke tests and rollbacks are part of deployments, improvements to caching layers for ci pipeline, removed serial requirement for ci pipeline.

**Other info**
This is a bit closer to a proper CI/CD pipeline.  Having a canary or QA deployment is a bit too much for us at the moment, however, we can have some serious confidence with the fact that unit tests and e2e tests are run as part of each build.  Additionally, cross-browser testing is done with a subset of tests as a way of testing the deployment, and will automatically roll back if issues are discovered.


lreading:
Initially I was trying to avoid pushing a new docker tag per commit, however it seems as though the caching solution implemented here isn't working exactly as I had hoped.  I think it's going to be best to just commit each hash and then separately maintain the latest tag, as well as versioned tags.   Will update.

lreading:
This will be merged via squash commit.  There's a ton of bad commits just getting the actions to trigger for testing. :sweat_smile:

Issue #60, opened 2021-03-19
Kindly suggest: Move the "close diagram" button


storkinsj:
**Describe what problem your feature request solves**
The close diagram button, a custom drawn "x" 2nd from the left in the edit pallet, is very difficult to discover as the "close diagram button". Normally this would be positioned all the way to the right (or maybe left on mac... but that's not an X). It is thicker than windows or XWindows style X's used for closing windows.  
    This so eluded our grasp that we thought we had to save and close the whole threat model file in order to access a different diagram... for more than 6 months until I tried clicking on the X. 
    
    I don't claim to be a UX expert or anything, but a smart team of 3 security engineers with a combined industry experience exceeding 70 years... could not figure it out. It's ok if you dont' fix this because we figured it out at this point. I know. RTFM.

**Describe the solution you'd like**
-Move the X all the way to the right
-Less delay on the fly-over help when the mouse is over the X
-Extra credit: Use the window close button and intercept the javascript; ask the user if he would like to close the diagram or the whole report. Then ask if they would like to SAVE first! :-)

**Additional Context**
LOL this drove us batty for months. But we finally did get it.

jgadsden:
Hello @storkinsj - this is a very good suggestion, thanks for raising it. The Threat Dragon core repo is the best place for this issue, would yuo like to move it to there?
https://github.com/OWASP/threat-dragon-core/issues

If you can provide a solution as well then that would be great - otherwise it will get changed in version 2.0 which should be out sometime late this year. Anyhow, thanks again for the suggestion.

lreading:
Threat Dragon version 2 is still in development, so the layout of the edit pallet may change before it's released.  That said, I think a good temporary solution is to simply highlight the main save/cancel actions by having the label next to the buttons as well:
![Screenshot from 2021-05-25 02-00-34](https://user-images.githubusercontent.com/5851985/119447085-5e504a00-bcfd-11eb-8915-39b35fe93141.png)

storkinsj:
Hi,
    The layout seems much clearer; however, is it a "cancel" button... or the "close" button? 
    It is normal to make edits in one diagram and close it then go to the appropriate other diagram. I think the word "cancel" will keep people from pressing it. I think the undo button is more about cancelling what you have done.
    
     By the way this does seem to belong in core. Should we move it there?

lreading:
[...]

PR #206, opened 2021-07-02
make return for diagram editor more obvious (issue #60)

jgadsden:
**Summary**
Issue #60 asks that the button to return from the diagram editor is made more obvious
The image has been changed to arrow-left, and moved to be on the far right. The delete and reload buttons have been moved to far left so that they are disassociated from the save and return

make return for diagram editor more obvious

**Other info**
move-return-button


jgadsden (commented):
Going ahead with these changes

PR #211, opened 2021-07-05
make boundaries black, removing colour

jgadsden:
**Summary**
boundaries are now black, so that the only colour is for components with unmitigated threats

remove colour to make boundaries black

**Other info**
looks a lot better, and is better for red/green colour sightedness
black-boundaries


jgadsden (commented):
very minor change, going ahead

Issue #150, opened 2020-06-05
Feature request: Clarify color


jgadsden:
This issue has been migrated from : 
https://github.com/mike-goodwin/owasp-threat-dragon-core/issues/75
and was opened by @adamshostack : 

The sample diagram has some elements (background worker, worker config, etc) which show in red. It's not clear what red means. (dashed lines seem to be used for both trust boundaries and Out of scope).

I suggest adding a key, but possibly alternately reducing use of color to address the threat of black & white printing.

lreading:
For v2, there will be a help section added to the diagram page itself that will show the indicators.  I agree with the OP that using colors isn't the best choice due to printers as well as people with color-blindness (green vs red can be very hard to differentiate).

Do we want to just include the key for v2, or should we discuss alternatives for UI indicators?

jgadsden:
Agreed, we want to move away from green and red.

My take on this is that  colour should not be used (solely) to indicate the status of any element, just as a further aid to what is already being indicated.

I guess we should fit in with what the generally generally expects, and I think it is only Threat  Dragon that provides dashed line borders to components that are out of scope - shame to lose that though

lreading:
I wouldn't want to lose the out of scope indicator.  Maybe we can standardize some icons that can be displayed along a line or within a box/circle that indicate different things?  The only downside of that is that it can quickly become cluttered looking.  We definitely want to balance the ability to quickly understand the threat model as well as keeping it readable.

Just spitballing, and this is off the top of my head so I may be missing some other important pieces of information.
|Icon/UI|Meaning|
|---|---|
|Triangle Exclamation|Has Threats|
|Open Padlock|Not encrypted (data flows)|
|Globe|Public Network|
|Small dashes|Out of Scope|
|Large Dashes|Trust Boundary|

These icons could even become part of the label?  This is related to #135 as well.

As mentioned elsewhere, I'm by no means a UI/UX expert. If anyone has suggestions, please do leave a comment!  I'm willing to try out as many different things as needed with the new Diagraming framework until we find a great solution. :smile: With that said, I'll try to put together a few different ideas and post the results here.  Sometimes it can be difficult to express thoughts on UI through words as opposed to pictures.

jgadsden:
![cisco_tb_example_web_app](https://user-images.githubusercontent.com/29352392/119946413-82768a00-bf8e-11eb-9d4e-d1bccbcdf5b7.jpg)

jgadsden:
[...]

PR #229, opened 2021-07-08
Removing the green color from trust boundaries per #150

lreading:
**Summary**
Removes the green color from trust boundaries, defaults to black.

Trust boundaries are black as seen in #150


jgadsden:
Cool, thanks for this @lreading , I did not mean to do a unilateral decision thing ... was just taking it out of the 1.4.x because it did not seem to add anything
If you want to keep colour for boundaries or make it configurable then that is OK as well, all good from my point of view

lreading:
I think it made perfect sense especially where the color coding is a common complaint, and where it's gone in 1.4.x, why not remove it from 2.0. :smile:

PR #233, opened 2021-07-10
return icon now left arrow not cross

jgadsden:
**Summary**
It seemed odd to have the  'cross' icon for when we return from an action. This was changed for the diagram view to a left arrow, and this pull request changes the 'Return' button on the report view to have a left-arrow as well

return icon now left arrow not cross

**Other info**


jgadsden (commented):
Small change, merging

Issue #195, opened 2021-06-23
Add N/A as Threat Level


lreading:
**Describe what problem your feature request solves**
Sometimes the "threat level" field is not applicable.  Currently, some users are using the mitigated property as a workaround, though that's a bit of a misnomer. 

**Describe the solution you'd like**
Add a "Not Applicable" option for the threat level.

**Additional context**
This suggestion was provided by Simon S

jgadsden:
I agree with this suggestion @lreading (and many thanks to Simon S for suggesting this). We have two options at present  'Open' and 'Mitigated' so we have a third in the same block? Suggest 'NA' for brevity's sake?

PR #246, opened 2021-08-01
Add NA as a Threat Level value

jgadsden:
**Summary**
Add NA to the Threat Level selector for STRIDE/LINDDUN/CIA threats

Add NA as a Threat Level value, issue #195 

**Other info**
This can also be forward ported to version 2.0


jgadsden:
Merging in as we need this for the version 1.5 release

PR #248, opened 2021-08-04
V2/diagram UI

lreading:
**Summary**
After discussions with end-users, it was uncovered that some external tooling may be consuming Threat Dragon's threat model JSON.  For that reason, we should not make any breaking changes to the JSON schema (only additive).

Reverting breaking schema changes

**Other info**
Thanks again to Simone S for the insight!


lreading:
@jgadsden - FYI: rolling back the breaking changes to the schema in v2 as a priority so that we don't get too far down the rabbit hole. :)

jgadsden:
Sounds good to me @lreading - if we are backwardly compatible then that should be OK. We added a UUID to the individual threats that does not break <1.5.0 so all should be good.

PR #251, opened 2021-08-04
V2/diagram UI

lreading:
**Summary**
Moving the temporary keyboard shortcuts listing on the graph page to a more production-ready location

Added a button on the graph ui to open a modal showing all shortcuts

**Other info**
Because the shortcuts include things you can do with the mouse, I intentionally named it "shortcuts" as opposed to keyboard shortcuts.  I think using the word "actions" may be confusing, because a user might think that an action has to do with the threat model.

The proposed implementation:
Button:
Screenshot from 2021-08-03 22-22-57

Modal:
Screenshot from 2021-08-03 22-23-04

### DIFF
--- a/td.vue/src/components/Graph.vue
+++ b/td.vue/src/components/Graph.vue
@@ -1,18 +1,6 @@
 <template>
   <div>
     <b-row>
-      <b-col>
-        <p>
-          Keyboard shortcuts: <b-badge>ctrl+c</b-badge> copy |
-          <b-badge>ctrl+v</b-badge> paste | <b-badge>ctrl+z</b-badge> undo |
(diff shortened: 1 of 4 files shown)


jgadsden (approved):
Looks good to me @lreading , nice to see this going in.

For darwin we may need a CmdOrCtrl + , for everything else ctrl + is perfect.

Possibly tripping point in the future is some MacOS conventions, which may need selection according to the OS
for example `close` is Cmd + W in MacOS world, but we are not implementing a `close` hare so all is good for now

lreading:
I left this open for a while because I didn't have my mac environment set up.  I just tested it and it appears the library is aware of the ctrl vs command paradigms for MacOS vs everything else.  Small update to the wording and I'll merge this. :)

Issue #250, opened 2021-08-04
Support github enterprise


danielpops:
**Describe what problem your feature request solves**
The github oauth configuration assumes that you're using github.com and doesn't work with github enterprise. 

**Describe the solution you'd like**
A configuration in .env (or equivalent for prod) that lets you specify a github enterprise url

lreading:
Thanks for the great suggestion, @danielpops !

This seems like it could be a quick win.  Threat Dragon uses [octonode](https://github.com/pksunkara/octonode) to connect to github, and it appears to support enterprise with relatively little config.

I don't have access to an enterprise account: if I put together a PR, would you be willing to test it to ensure it's working as expected?

jgadsden:
I am just about to release version 1.5.0 - delayed because of problems with electron-builder config file - so depending on how quick this win is it can go in 1.5.0 or if it is a bit slower we can do a version 1.5.1

lreading:
@danielpops - if you have a chance, can you test out this branch? https://github.com/OWASP/threat-dragon/tree/feature/enterprise-github-config-250

I added config options for a Github enterprise hostname, and optionally port/protocol.  The only update you should need to test the change and the enterprise integration is adding `GITHUB_ENTERPRISE_HOSTNAME` to your env file

danielpops:
@lreading hey, this seems to have *kind of* worked. I was able to authenticate through my github enterprise / sso credentials. However, the app isn't able to view the repos, with a `Bad credentials` error messgae

```
{
    "hostname": "e2c837b09ef2",
    "level": 50,
[...]

PR #252, opened 2021-08-04
Adding support for github enterprise *needs testing*

lreading:
**Summary**
Adding support for Github enterprise.  

#250 - Adds support for Github enterprise by adding the hostname (optionally port and protocol) to the config

**Other info**
This needs to be tested by someone who has access to a Github enterprise account.  I will review docs and see if any updates need to be made there.


lreading:
Thanks for taking a look Jon! To be clear, my intention is to leave this PR open until we can have it tested by someone with access to an enterprise account. I'm relatively confident that we're not introducing new bugs or regressions, but I'm not comfortable merging this and claiming we support it without someone testing first. :smile:

jgadsden:
I created a free 16 day github enterprise account. It did not help with testing though - I used the cloud hosted offering and this results in a github.com host URL ... so no benefit to us as this is the same as the free account.

We would need a github enterprise account for our own servers to test this. I checked with the ForgeRock enterprise account and we also use the cloud hosted version - so again it has a github.com host

lreading:
Funny, I was looking at the cloud enterprise trial earlier.  That's unfortunate.  I'll shoot a message out in the slack channel tomorrow to see if we have any enterprise users in there that would be willing to lend a hand in testing. :)

jgadsden (approved):
@danielpops and @lreading , should we now retire this PR? We can then start afresh on  the v2-development branch

lreading:
Closing - Will try again in the v2 codebase (with better logging/output).

PR #255, opened 2021-08-10
Remove unnecessary `git init` step from Installing section

gabe-sky:
This commit removes the `git init` step, which is unnecessary and potentially
dangerous if done in, e.g., a user's Documents directory.

It also updates the documentation to tell the user where the clone will go.

(It also fixes a typo in the description of the web application.)

**Summary**
This updates the installation instructions so that users don't end up
running `git init` somewhere they don't want to.  If they did so, it
turns the current directory into a git repository.

Remove unnecessary `git init` step from Installing section

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The application comes in two variants:
 
 1. [**A web application (this repo)**](https://github.com/owasp/threat-dragon):
 For the web application, models files are stored in GitHub (other storage will become available).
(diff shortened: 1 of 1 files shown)


lreading (approved):

jgadsden (approved):

jgadsden:
merging this in - many thanks for the fixup @gabe-sky

Issue #281, opened 2021-09-24
td.site tests are disabled - needs reverting


jgadsden:
**Describe the bug**
bug fix #280 was a blocker and needed to be fixed. The tests were throwing an error which was stopping the fix.

The td.site testing had to be disabled, the app works on all my browsers, and this is the error on tests :

```
24 09 2021 09:45:30.467:ERROR [framework.browserify]: bundle error
24 09 2021 09:45:30.470:ERROR [framework.browserify]: SyntaxError: 'import' and 'export' may appear only with
'sourceType: module' (1056:0) while parsing /Users/jon.gadsden/owasp/bugfix-threatengine-error/td.site/node_modules/jsonpath-plus/dist/index-browser-esm.js
while parsing file: /Users/jon.gadsden/owasp/bugfix-threatengine-error/td.site/node_modules/jsonpath-plus/dist/index-browser-esm.js
```

**Expected behaviour**
fix the tests and re-enable them

**Environment**
 - Version: latest
 - Platform: Web App only
 - OS: any MacOS / Windows / Linux
 - Browser: any chrome, firefox, safari

**To Reproduce**
run `npm test` from the top level directory

**Any additional context, screenshots, etc**
[This commit](https://github.com/OWASP/threat-dragon/pull/280/commits/7d7fcc3b44e20bf2a41fd20d842ecf84f796081c) is the temporary fix that needs to be reverted.

```
  -  "test": "npm-run-all build pretest",
  +  "test": "npm-run-all build pretest test:firefoxheadless",
```

lreading:
I've narrowed this down to jsonpath-plus, which is a dependency of json-rules-engine, which is a dependency in core.  I can't quite sort out why the browserify process is choking on it.  I'll keep poking at it for now, but if anyone is experienced with transpilation issues like this, any help would be appreciated! :smile:

lreading:
The simplest, though not the "best" solution that I've found is to update the karma config to load the webpack-bundled threatdragon javascript.  This bundle is generated using the `npm run build` command, which is run when you run `npm run test`.  This isn't a perfect solution, since you can run any of the `npm run test:xxx` commands and not update the bundle first, thus not get accurate output.  We are also losing out on some of Karma's other nicer features like watching and whatnot.

I'm sure there's a way of syncing the two up to work together, but I just don't really know where to start with that. :) I'm opening a PR with the hacked version for discussion.

jgadsden:
Many thanks @lreading for sorting this out

PR #282, opened 2021-09-25
Enable td.site Tests 

lreading:
**Summary**
Enables td.site tests

Re-enabling td.site tests.  For td.site, please only use `npm run test` moving forward, or you may not be testing the most up-to-date code

**Other info**
Discussion on implementation in #281

### DIFF
--- a/td.site/karma.conf.js
+++ b/td.site/karma.conf.js
@@ -14,11 +14,11 @@ module.exports = function (config) {
 
     // list of files / patterns to load in the browser
     files: [
-      { pattern: 'node_modules/babel-polyfill/browser.js', instrument: false},
-      'src/app/app.js',
-      'node_modules/angular-mocks/angular-mocks.js',
-      'test/**/*.js',
-      'src/app/**/*.html'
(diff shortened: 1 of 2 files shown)


lreading:
hm, not sure what's going on with that travis ci build.

jgadsden:
@lreading I think this is a random error with npm install - so I am rerunning the checks ...

jgadsden (approved):
Looks good to me - not sure whay the checks failed

@lreading I think this is a random error with npm install - so I am rerunning the checks ... hmmm, I can not do that - maybe because it is not my PR. Did you want to try rerunning?

jgadsden:
And just to confirm that when I ran and tested this locally it worked correctly, so this really does seem to be a random Travis error

lreading:
It is also not allowing me to re-run the install. V2 we are using GH Actions exclusively for the build.  We could move this branch over to actions as well if we wanted. It would make it easier to re-run failed jobs?

I'm going to merge this. I agree that it's the random build error. :)

PR #287, opened 2021-09-29
Plumbing In Model Creation

lreading:
**Summary**
Plumbing in the ability to add new models

The new model button now works, though the back-end is not hooked up yet.

**Other info**
Back-end will be covered in other work


lreading:
Converting to draft until I add e2e tests.  Creating a new model and using the edit page would likely make a great smoke test as well.

PR #290, opened 2021-09-30
V2 - Adding ability to import model via JSON text in local session

lreading:
**Summary**
Local sessions need a way of importing threat models.  One way that helps with development and debugging is allowing for copy/pasting of the raw data into the UI.

Local sessions can now import JSON text directly.  This is accomplished via an additional action menu item on the dashboard after selecting local session.

**Other info**
This may not be the final implementation or even make it to the official v2 release.  Feedback is welcome.


lreading:
@jgadsden - FYI
I think this feature will be helpful for getting users to give feedback before more features are ready.  For now, that would be limited in scope to only the diagram UI, minus threats and threat generation.  More on comparing v1.x to v2 and ensuring models are loading as expected.

PR #298, opened 2021-10-20
V2/server logging refactor

lreading:
**Summary**
Logging refresh for td.server.  The server application should be given a once-over to review the existing logs and be made a bit more consistent/helpful.

Refactor v2 td.server logging to use winston

**Other info**
The following logging levels are available, with the lower number being the highest priority:
    audit: 0,
    error: 1,
    warn: 2,
    info: 3,
    debug: 4,
    silly: 5

By default, 2 log files will be created: app.log and audit.log.  app.log will log everything at and above the level specified in your env.  The audit log will only log audit entries, and could (in theory) and was separated for ease of ingestion via other security tools.

### DIFF
--- a/.github/workflows/.zap-rules-web.tsv
+++ b/.github/workflows/.zap-rules-web.tsv
@@ -7,6 +7,8 @@
 10062	OUTOFSCOPE	.*_bom\..*
 10094	OUTOFSCOPE	.*_bom\..*
 10110	OUTOFSCOPE	.*jquery\.min\.js
+10094	IGNORE	Base64 Disclosure
+10027	IGNORE	Information Disclosure - Suspicious Comments
 10096	IGNORE	Timestamp Disclosure - Unix
 10099	IGNORE	Source Code Disclosure - SQL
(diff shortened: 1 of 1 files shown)


lreading:
> This pull request **introduces 1 alert** when merging [7652dd6](https://github.com/OWASP/threat-dragon/commit/7652dd618ea3e7c2f8f74c32f12e5b7ea49a62ca) into [222b3b1](https://github.com/OWASP/threat-dragon/commit/222b3b1155cbd78a4810b5da9f23880da522d977) - [view on LGTM.com](https://lgtm.com/projects/g/OWASP/threat-dragon/rev/pr-05d9318ec01fe4600affc819cb20d2e819e753f4)
> 
> **new alerts:**
> 
>     * 1 for Syntax error

It's not happy about the private method in the class, which I _believe_ is still experimental anyways.  I'll also have to look into the new zap failures as well.  I did add a new dependency (winston), so that may have to do with it.  Will fix these before merging.

PR #316, opened 2021-11-09
Providing translation from English to French

andk123:
**Summary**
Provides translation from English to French.

#297 - Adding EN to FR translation

**Other info**
PR inspired from #315 🙂

### DIFF
--- a/td.vue/src/i18n/fr.js
+++ b/td.vue/src/i18n/fr.js
@@ -1,0 +1,201 @@
+const fr = {
+    auth: {
+        sessionExpired: 'Votre session est expirée. Veuillez vous reconnecter pour continuer.'
+    },
+    nav: {
+        v2Warning: 'Voici la version 2.0 de OWASP Threat Dragon et elle est en cours de développement. Ne l\'utilisez pas pour modifier des modèles existants, car cette version pourrait les briser!',
+        loggedInAs: 'Connecté en tant que'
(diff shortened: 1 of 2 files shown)


andk123:
@lreading I inspired myself from your previous PR. Let me know if I'm missing anything!

jgadsden:
Good to see @andk123 

We have more text in the proposed threat engine, but we need to discuss what can be done about that . Did you want ot have a look at the threat engine PR and see how feasible a translation would be for `threatengine.js` ?

lreading (approved):
Looks great.  Thank you again for taking the time to do this!!!!  Would you like the honor of clicking the merge button? :)

andk123:
No problem at all! 👍  @lreading 

& @jgadsden I just had a look, the translation for threatengine.js would be just fine however, I might need some help with the structure of the fr.js for the threat engine. It was handled here using Vue.js, but I'm not sure what would be the best way to handle it in a vanilla js file

jgadsden:
> I just had a look, the translation for threatengine.js would be just fine however, I might need some help with the structure of the fr.js for the threat engine. It was handled here using Vue.js, but I'm not sure what would be the best way to handle it in a vanilla js file

Hello @andk123 - yes, totally agree, I meant to say that we need to look at the threat engine tranlsation at some point, but not for this particular PR

PR #317, opened 2021-11-20
Update README.md dockerhub example to match recommendation

hwine:
Have dockerhub example use the "stable" version as suggested

**Summary**

**Other info**

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,8 @@ Threat Dragon maintains a docker image on Dockerhub.  Each release is tagged as 
 
 Alternatively, you can use the `stable` tag, which will always be the latest official release.
 
-- `docker pull threatdragon/owasp-threat-dragon:v1.5.8`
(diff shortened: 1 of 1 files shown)


lreading (approved):
Originally I thought it would be good to keep the specific version tag, as I think that's how most people will wind up deploying it in their environment. I think it does make more sense to use the stable tag in the readme for a few reasons, one of which being that it matches the dockerhub example as you've brought up here.  Another is that we won't need to update the readme each time we release, which is just one less thing to forget later. :grin: 

Thanks for raising the issue and opening the PR! Feel free to merge!

hwine:
> Thanks for raising the issue and opening the PR! Feel free to merge!

I can't merge, as I don't have (or need) write permissions in the repo! 😀 This was just a drive-by doc update (though I will be watching the project).

PR #322, opened 2021-12-05
V2/i18n el

kostasadriano:
**Summary**

**Other info**


kostasadriano:
This was my first pull request and I probably made a mess with the commits, creating/deleting files and updating the same file 5-6 times... If you prefer, I can create my branch again and resend the pull request.

jgadsden:
Superb, thankyou @kostasadriano for this Greek translation

Do not worry about the multiple commits - we can squash when merging into main branch, which will result in just  the one commit message

jgadsden:
We need to update file `td.vue/src/i18n/index.js` for the selection of 'el'. This is present on v2-development branch, and not on OWASP:main

kostasadriano:
Ok, I'll make a new pr.


> We need to update file `td.vue/src/i18n/index.js` for the selection of 'el'. This is present on v2-development branch, and not on OWASP:main
> 
> @kostasadriano do you want to resubmit this pull-request, comparing it against 'v2-development' rather than 'main'? File `td.vue/src/i18n/index.js` can then be fixed up for the selection of 'el', along the lines of:
> 
> ```
> import Vue from 'vue';
> import VueI18n from 'vue-i18n';
> 
> import en from './el.js';
> import en from './en.js';
> import es from './es.js';
> import fr from './fr.js';
> import pt from './pt.js';
> 
> Vue.use(VueI18n);
> 
> let i18n = null;
> 
> const get = () => {
>     if (i18n === null) {
>         i18n = new VueI18n({
>             locale: 'en',
>             messages: {
>                 el,
>                 en,
>                 es,
>                 fr,
>                 pt
>             }
>         });
>     }
>     return i18n;
> };
> 
> export const tc = (key) => get().tc(key);
> 
> export default {
>     get
> };
> ```

PR #335, opened 2021-12-31
Add scrolling

### LINKED ISSUE #331: Version 2.0: add scrolling
At the moment we can not scroll across the canvas. We can zoom out and place elements in parts of the canvas that are not accessible when we zoom back in

It would be good to be able to scroll the canvas, ether by dragging an element or bumping the edge of the visible area of the canvas. Ideally the cursor / mouse pointer should stay within the visible area of the canvas

**Additional context**
This is probably already provided by antv/x6, and just needs enabling

lreading:
**Summary**
Adds scrolling to the graph

Closes #331 
Mousewheel works as expected.  Holding shift and using mousewheel will scroll horizontally.

### DIFF
--- a/td.vue/src/service/x6/graph/graph.js
+++ b/td.vue/src/service/x6/graph/graph.js
@@ -66,6 +66,9 @@ const getConfig = (container) => ({
     connecting: {
         allowNode: true,
         allowBlank: true
+    },
+    scroller: {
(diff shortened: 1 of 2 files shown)


jgadsden (commented):
Tried it out and it works a treat - very nice to see

PR #347, opened 2022-01-05
proxy env var now called PROXY_PRESENT

jgadsden:
**Summary**
rename env var PROXY as PROXY_PRESENT 

@taoyiyang1 reported that we have problems with certain proxy configurations. An environment variable PROXY is used to correct this, and this variable is now named PROXY_PRESENT to fit better with the other env var naming. Also documents the default value

**Other info**
Hope this is OK with you @taoyiyang1 ?


lreading (approved):
Travis is reporting test failures, other than that makes sense and looks great!!!

jgadsden:
Thanks @lreading , I had pushed the updated test file to the wrong branch, all working now

taoyiyang1 (commented):
Sure, I think it's great

PR #348, opened 2022-01-05
V2/339 report

### LINKED ISSUE #339: Version 2.0 - Report Feature
Version 2.0 should have a reporting feature with similar outputs to version 1.x

- The report button generates a report
- Checkbox options for:
  - Show out of scope elements
  - Show mitigated threats
  - Include threat model diagrams
- Buttons at top/bottom (maybe sticky in the top-right for usability?)
  - Save PDF
  - Print
  - Return

**Additional context**
Some Threat Dragon users rely heavily on the reports, and some are still printing them.  The reporting feature is a must-have for the initial 2.0 release.

The Save PDF button will be implemented in #342

lreading:
**Summary**
Implements the reporting feature, including the web-based report and a printed report.  PDF is a TODO.

Implement reporting feature for web and printers
Closes #339 

**Other info**
This is a draft for feedback.  
Known bugs / workdarounds:
- ~You need to _edit_ the diagram before going to the report page.  Currently the edit window is "upgrading" the diagram to the expected v2 entities.  This will be fixed prior to merging~

TODO:
- ~Unit tests for `views/Report.vue`~
- ~End to end tests for report / printer report~
- ~Address bug (upgrade version logic) mentioned above~

I opted against writing an e2e test for the printer report.  It shares a lot of code with the report page, and is intended to be viewed from a native print dialogue.  This does not lend itself well to testing, however, we can circle back to it if others think there's value in having testing around it.


lreading:
This PR also closes the gap on unit test coverage. While I don't expect to be able to maintain 100% test coverage, it's nice to have a solid foundation for future development effort.

lreading:
The upgrade version logic was getting a bit messy and needed in multiple places.  In an attempt to solve this, I made an "upgrade" page that will show the user a modal explaining that v2 is using a different drawing lib, and that there _could_ be some slight differences in their models' appearances.  It shows all the models for the specific diagram.

This is implemented by adding it as a router guard, checking against the currently loaded threat model.  If the model is not tagged as version 2, it will bring the user to this upgrade page.  After this upgrade, the user will not see this page again and can continue as normal.

Screenshot from 2022-01-06 00-26-57

lreading:
Ready for review

jgadsden (approved):
Looks good to me, a huge amount of great work. Just a couple of observations from me:

The 'out of scope' column for the individual elements in the report takes up space - and has no extra meaning when the 'Show out of scope elements' is not ticked. Could this be done differently? Though I am not sure how

It seemed odd to go to the report page and then on to the printer-friendly view, especially as the tick box status did not carry over. It seemed better to go straight to the printer-friendly view

lreading:
Both great points!

I'll remedy the report/print.  You're 100% correct that a user shouldn't care if the report is "printer friendly" or not, as we can manage that independently.  For development, it was significantly easier to write two separate reports for formatting, but realistically, we can easily hide the printer version when the media is screen, and vice-versa.  I'll make those changes. :smile: 

Regarding out of scope elements, I see at least a couple of solutions:
- Remove the out of scope column entirely when the checkbox is un-ticked.  Replace the words "out of scope" with a checkmark or similar in the table data when it is being shown.
- Add an indicator at the entity level, eg "Browser - (Actor/Out of Scope)".  This risks losing context if the table data spreads to a new page.

To me, what makes it a bit more difficult is that out of scope is defined at the entity level, not the threat level.  From a user perspective, however, I still care about that information when looking at threats. I'm not 100% sure how to handle it either. :)

lreading:
Changes made:
- Report is now controlled by the media query (screen vs print)
- No more "printer friendly" button, just the "print" button
- OutOfScope is reported only at the entity level and no longer at the threat level.  We can re-evaluate in the future based on feedback if necessary.  This means there is no "out of scope" column in the threat data tables.

This does feel quite a bit cleaner/polished.  Thanks again, as always, @jgadsden  for the great feedback!

If there are any issues or features needed, we can address those as separate issues.  I'll be merging this once CI is complete so I can move on to other work. :smile:

PR #351, opened 2022-01-08
The first Sinicization

### LINKED ISSUE #350: Add internationalisation for China
We have various language translations available for version 2.0 Threat Dragon. It would be good to have a translation for China as well

Provide the translation, similar to the other translations under `td.vue/src/i18n/*.js`, and then fix up `td.vue/src/i18n/index.js` to read something like:

```
import Vue from 'vue';
import VueI18n from 'vue-i18n';

import cn from './cn.js';
import el from './el.js';
import en from './en.js';
import es from './es.js';
import fr from './fr.js';
import pt from './pt.js';

Vue.use(VueI18n);

let i18n = null;

const get = () => {
    if (i18n === null) {
        i18n = new VueI18n({
            locale: 'en',
            messages: {
                cn,
                el,
                en,
                es,
                fr,
                pt
            }
        });
    }
    return i18n;
};

export const tc = (key) => get().tc(key);

export default {
    get
};
```

**Additional context**
A similar process was for a translation to Spanish in #319

taoyiyang1:
**Summary**
What existing issue does the pull request solve?
Add Internationalization for China

Add cn.js file, modify i18n/index.js
fixes #350

### DIFF
--- a/td.vue/src/i18n/cn.js
+++ b/td.vue/src/i18n/cn.js
@@ -1,0 +1,205 @@
+const cn = {
+  auth: {
+      sessionExpired: '会话登录已过期，请重新登录。'
+  },
+  nav: {
+      v2Warning: '这是 OWASP Threat Dragon 的 2.0 版本，仍在开发中。不要使用它来编辑现有模型，因为这个版本可能会破坏它们！',
+      loggedInAs: '登录身份为'
+  },
+  home: {
+      title: 'OWASP Threat Dragon',
+      imgAlt: 'Threat Dragon Logo',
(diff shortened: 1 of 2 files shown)


jgadsden (approved):
Superb, thankyou @taoyiyang1 , this looks great to me - I wandered around the application in Chinese and it all seems good.
I see that some text has remained in English, and I guess this is because written Chinese does not translate these words and the English is used.
@lreading the fact that the app works perfectly well with Chinese characters and Greek characters is a testament to the I18N you put in place.

jgadsden:
@taoyiyang1 do you want to have the honour of merging this in?

taoyiyang1:
> 太棒了，谢谢@taoyiyang1，这对我来说看起来很棒 - 我在中文应用程序中徘徊，一切看起来都很好。 我看到有些文字仍然是英文的，我猜这是因为书面的中文没有翻译这些单词，而是使用了英文。 @lreading应用程序与中文字符和希腊字符完美配合的事实证明了您使用的 I18N。

@jgadsden Yes, some places have not been translated for the sake of accuracy. I will think about it later and see if I need to translate it.

lreading:
There will probably be new translation strings added later, so do not stress over some things not being 100% done! :) As you can tell, the v2 branches are very active and changing frequently.  Thank you so much for doing this! Having a Chinese translation will be huge for our user base!

PR #362, opened 2022-02-05
feat: wrap vue web app in electron

### LINKED ISSUE #334: v2 - desktop version
We should have a desktop version in place before version 2.0 release

It would be good to use an electron wrapper for the web application, and this will provide the desktop version of the threat dragon application

**Additional context**
Note that the existing version 1.x threat dragon desktop uses common code with the web application, but does not wrap the web application itself.  This has caused an increase in development and test effort, so it would be good to change the way this is done. Refer to [this example](https://jojozhuang.github.io/tutorial/converting-web-app-to-desktop-app-with-electron/) but it can be done in various ways

ShubhamPalriwala:
Signed-off-by: ShubhamPalriwala <spalriwalau@gmail.com>

**Summary**

Wraps the existing Vue web app using the `electron-builder` vue [plugin](https://nklayman.github.io/vue-cli-plugin-electron-builder/).

**Steps Used:**

```sh
pnpm install vue-cli-plugin-electron-builder
vue add electron-builder
yarn electron:serve (To start dev mode)
```

**Demo:**
[Demo Video](https://streamable.com/e3aq1n)

Fixes #334


jgadsden:
Hello @ShubhamPalriwala - looks good and I will checkout the code and run it
Your editor has reformatted the package.json to have indents of 4 spaces, but we want to keep to indents of 2 spaces - could you go back to 2 space indent for this file please? Thanks - and I will have a look at the desktop version

ShubhamPalriwala:
Oh! Sorry for the trouble! I've pushed another commit with the fix in tab size. Lmk if you want me to squash both of them into 1 when you have tested it.

You can test the the feature using:
```
yarn electron:build: To build the app (Production),
yarn electron:serve: To run the app (Development),
```

lreading:
This is looking great, thanks so much for tackling this one @ShubhamPalriwala !

I think it'd be best if we were to remove the need to run `vue add electron-builder`.  I didn't read your clearly written instructions before attempting to run it, and ran into an error. :sweat_smile: I was able to `electron:serve` (without running the vue cli command) by adding `electron-builder` as a dev dependency.  This would just eliminate a step for anyone who tries to build/run locally in the future.  I don't think we would want to utilize a pre-install script or similar, as that might add some overhead to the docker builds.  This is not critical, just my thoughts!  

I tested on Ubuntu, and I'll have a check on Mac/Windows in the next day or two just to confirm.  I wouldn't let that hold this up though, as it's a new feature and we can raise bugs if anything breaks.


Again, _awesome_ work @ShubhamPalriwala!

td.vue/package.json
```diff
@@ -58,6 +60,8 @@
     "browserstack-cypress-cli": "^1.8.1",
     "chromedriver": "^90.0.0",
     "codecov": "^3.8.2",
+    "electron": "^13.0.0",
+    "electron-devtools-installer": "^3.1.0",
```
jgadsden, line 64: Could we use the latest version of electron-devtools-installer which is 3.2.0 ?

td.vue/package.json
```diff
@@ -58,6 +60,8 @@
     "browserstack-cypress-cli": "^1.8.1",
     "chromedriver": "^90.0.0",
     "codecov": "^3.8.2",
+    "electron": "^13.0.0",
```
jgadsden, line 63: The latest  version of electron is 17.0.0 - I have tried it with this version and it seems to work OK

td.vue/package.json
```diff
@@ -73,9 +77,11 @@
     "portal-vue": "^2.1.7",
     "sass": "^1.32.12",
     "sass-loader": "^10.1.1",
+    "spectron": "15.0.0",
```
jgadsden, line 80: latest version of spectron is 19.0.0, can we use this latest version?

td.vue/package.json
```diff
@@ -20,9 +24,7 @@
     "lint": "vue-cli-service lint",
     "codecov": "codecov"
   },
-  "resolutions": {
-    "ansi-regex": "^5.0.1"
-  },
+  "main": "background.js",
```
jgadsden, line 27: if we rename away from the default `background.js` then this line may need to be updated

td.vue/src/background.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
```
jgadsden, line 1: can we rename this away from the default `background.js` ?
something with `electron` and `td` in it would indicate what the file is used for ... not worried what the name is so long as it is clear

jgadsden (commented):
Hello @ShubhamPalriwala - this is a great first step. I had some comments about the default names / package revisions, if you could have a look at them.

The electron  tests did not  run for me:
```

 FAIL  tests/unit/electron.spec.js (55.769s)
  ✕ Window Loads Properly (50026ms)

  ● Window Loads Properly

    : Timeout - Async callback was not invoked within the 50000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 50000ms timeout specified by jest.setTimeout.Error:

       6 | jest.setTimeout(50000);
       7 |
    >  8 | test('Window Loads Properly', async () => {
         | ^
       9 |     // Wait for dev server to start
      10 |     const { app, stopServe } = await testWithSpectron(spectron);
      11 |     const win = app.browserWindow;

      at new Spec (node_modules/.pnpm/jest-jasmine2@24.9.0/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Object.<anonymous> (tests/unit/electron.spec.js:8:1)
```

jgadsden (approved):
I have opened a discussion page for the V2.0 electron feature:
https://github.com/OWASP/threat-dragon/discussions/363

This pull request is a great step towards that feature, so gets my appproval and `i suggest we go ahead and merge 
Requesting review also by @lreading

jgadsden:
I think this is only wrapping td.vue - we also need td.server. I can look into this if you like @ShubhamPalriwala ?

lreading:
> I think this is only wrapping td.vue - we also need td.server. I can look into this if you like @ShubhamPalriwala ? I think others have the same problem : https://stackoverflow.com/questions/69322243/electron-building-issue-for-server-along-with-the-front-end
> 
> when running command `v2-electron/td.vue$ yarn electron:serve` and accessing github we get
> 
> ```
> Proxy error: Could not proxy request /api/login/github from localhost:8080 to http://localhost:3000.
> See https://nodejs.org/api/errors.html#errors_common_system_errors for more information (ECONNREFUSED).
> ```
> 
> If I first run the server from the command line with `v2-electron$ npm run dev:server` then the `v2-electron/td.vue$ yarn electron:serve` can access github as expected

My original plan was to add some wrapping to the auth methods to lock it down to local only when it's running in electron.  I do have a selector in td.vue somewhere to determine if it's running in an electron instance, it's just un-tested and not implemented yet.

We don't _have_ to do it that way though, we could wrap the server in electron as well.  I did some really light research on it a while ago, but haven't ever actually done it.


All that said, that feels like work for another day.  This is a huuuuuuge step forward and great start. <3

lreading (approved):
Nothing to add here, I think Jon raised some good points as well.  Thanks again for the great work!

ShubhamPalriwala:
> 



> This is looking great, thanks so much for tackling this one @ShubhamPalriwala !
> 
> I think it'd be best if we were to remove the need to run `vue add electron-builder`. I didn't read your clearly written instructions before attempting to run it, and ran into an error. sweat_smile I was able to `electron:serve` (without running the vue cli command) by adding `electron-builder` as a dev dependency. This would just eliminate a step for anyone who tries to build/run locally in the future. I don't think we would want to utilize a pre-install script or similar, as that might add some overhead to the docker builds. This is not critical, just my thoughts!
> 
> I tested on Ubuntu, and I'll have a check on Mac/Windows in the next day or two just to confirm. I wouldn't let that hold this up though, as it's a new feature and we can raise bugs if anything breaks.
> 
> Again, _awesome_ work @ShubhamPalriwala!

I am totally in for anything that makes the user's life easier. I have added the `electron-builder` as a dev dependency and tested the same. Thanks for the quick feedback!

td.vue/package.json
ShubhamPalriwala: Done!

td.vue/package.json
ShubhamPalriwala: Removed this line! Thank you for your suggestion.

ShubhamPalriwala:
> Hello @ShubhamPalriwala - this is a great first step. I had some comments about the default names / package revisions, if you could have a look at them.
> 
> The electron tests did not run for me:
> 
> ```
> 
>  FAIL  tests/unit/electron.spec.js (55.769s)
>   ✕ Window Loads Properly (50026ms)
> 
>   ● Window Loads Properly
> 
>     : Timeout - Async callback was not invoked within the 50000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 50000ms timeout specified by jest.setTimeout.Error:
> 
>        6 | jest.setTimeout(50000);
>        7 |
>     >  8 | test('Window Loads Properly', async () => {
>          | ^
>        9 |     // Wait for dev server to start
>       10 |     const { app, stopServe } = await testWithSpectron(spectron);
>       11 |     const win = app.browserWindow;
> 
>       at new Spec (node_modules/.pnpm/jest-jasmine2@24.9.0/node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
>       at Object.<anonymous> (tests/unit/electron.spec.js:8:1)
> ```

I'll take a look at this and push a fix for the tests. Thank you for pointing this out, really sorry I missed it.

td.vue/src/background.js
ShubhamPalriwala: Agreed! How about renaming this to `td.desktop.js`?

jgadsden: Yes, sure - `td.desktop.js` sounds good to me

td.vue/package.json
ShubhamPalriwala: Done!

td.vue/package.json
ShubhamPalriwala: Done! Thank you

td.vue/src/background.js
ShubhamPalriwala: Done! Thank you for this suggestion too!

ShubhamPalriwala:
Pushed a commit w all the suggestions incorporated till now. Only thing left is the Electron test issue for which I have opened an issue as I saw a lot of people facing the same issue but none of their solutions worked for us.
Waiting for the response for it [here](https://github.com/nklayman/vue-cli-plugin-electron-builder/issues/1657)

ShubhamPalriwala:
If you are fine with adding the test file later then I can remove the failing test file and then we can merge it!

jgadsden:
All done, thanks again @ShubhamPalriwala

ShubhamPalriwala:
Hey @jgadsden We had to remove the failing test case file :((
Should I open another PR removing that file? 
Because right now the entire tests suite will fail due to this one test case

jgadsden:
No problem - you should have write access? It will make sure we fix it ... I would rather fix it than remove it? I can look at getting the tests running this morning

ShubhamPalriwala:
Thank you so much! I'm waiting for the update here https://github.com/nklayman/vue-cli-plugin-electron-builder/issues/1657.

See if you can do anything!

PR #367, opened 2022-02-09
Adding EOL notice for v1 in README

lreading:
**Summary**
Adds a notice about v1 EOL and where to find more information on running v1.x

Don't be afraid of providing feedback on this one! I just wanted to get the discussion/idea started, if there's a better way I am 100% behind it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ Each release provides installers for Windows, Mac OSX and Linux.
 
 [End user help](https://www.threatdragon.com/docs) is available for both variants.
 
+
+### Version 1.x End Of Life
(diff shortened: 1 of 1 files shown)


ShubhamPalriwala:
I like the idea! Can we also define the tab spaces and basic formatting rules so that we do not have to face any formatting problems!

jgadsden (approved):
Nicely put, gets my  approval

jgadsden:
I agree @ShubhamPalriwala that it would be good to have some guidelines for the coding style, at the moment we tend to follow the style of the file that we are changing. Did you want to raise an issue for this?

lreading:
I think a separate issue would be great.  We can also tailor the eslint rules and .gitconfig for line endings, indentation, code style, etc. :)

ShubhamPalriwala (approved):
Lgtm! All hail v1

PR #376, opened 2022-02-12
Implement electron specific providers

### LINKED ISSUE #373: v2 - desktop includes github login
**Description**
When using the desktop application, if the main page button is pressed 'Login with GitHub' then the main window becomes the github login screen and there does not seem to be a way to escape

**Expected behaviour**
It would be good to be able to get back to the main screen without having to close / reopen the application

**Environment**
 - Version: 2.0
 - Platform: Desktop App
 - OS: all , but tested on MacOS
 - Browser: NA

**To Reproduce**
1. from command line `npm run start:desktop`
2. click on 'Login with github' button
3. there seems to be no escape from the github login window

Note that if using the installed desktop app (rather than one run via vue cli) then there is different bahaviour

attached screenshot shows the github login screen 
threat-dragon-github-login

lreading:
**Summary**
For v2.0, only "local session" will be available as a login option in the electron version of Threat Dragon.

Implements logic to detect when running in electron, modifies the list of available login providers based on that.
Closes #373 

Browser:
Screenshot from 2022-02-12 13-37-28

Electron:
Screenshot from 2022-02-12 13-38-05


lreading:
Some of the style changes were done by running `pnpm run lint` which has the `--fix` flag applied.  I know Shubham raised an issue on making the styles and such standardized, maybe we can add linting as a pre-commit command as part of that. :)

ShubhamPalriwala (approved):
Lgtm! Another small but effective step forward! Lessgooo

PR #379, opened 2022-02-13
Resolving vulnerable dependencies in td.vue, as picked up by Trivy

lreading:
**Summary**
The latest push to main flagged a vulnerable library in our production image.

Removes vulnerable versions of node-fetch and follow-redirects

**Other info**
We should consider running trivy on PRs as well (if it's not already).  It could also have been a timing issue where the vulnerabilities were disclosed between when the PR was open and when it was merged.  I will look into this later.


lreading:
The CI pipeline isn't running on PRs... Taking a look!  I'll also take a swing at the browserstack nightly pipeline that failed today as well

.github/workflows/ci.yaml
```diff
@@ -3,6 +3,10 @@ on:
   push:
     branches:
       - '*'
+  pull_request:
+    branches:
+      - 'main'
+
```
ShubhamPalriwala, line 9: Regarding the Trivy scan, it currently scans the master branch on docker! We change this to a filesystem scan so that it locally downloads the files and does a local file scan on another PR branch instead!

If you want, I can work on it and raise a separate PR

ShubhamPalriwala (commented):
Rest all apart from Trivy CI Scan suggestion lgtm!

lreading:
Trivy is still unhappy, I may have jumped the gun requesting review.  Looking a bit deeper!

.github/workflows/ci.yaml
ShubhamPalriwala: Oh alright! That makes sense! I like the dev dependencies being skipped for now!
Everything in the PR Lgtm now!

ShubhamPalriwala (approved):
We just need to fix the Trivy Scan prolly need to update a few more deps! Lgtm

lreading:
> We just need to fix the Trivy Scan prolly need to update a few more deps! Lgtm

Same dep was in td.server (which already had an override for 1.14.7, same as td.vue did) :face_exhaling: 
I'm building and testing locally _first_, then will push the update :sweat_smile:

.github/workflows/ci.yaml
lreading: If you think there's value in doing it another way, by all means please do open an issue!  I'm not saying this way is 100% correct or the absolute best/perfect way of doing it, just explaining why it was configured that way.  I'm all about improving things!

PR #382, opened 2022-02-13
Feature/340 threatmodeledit save

### LINKED ISSUE #340: v2 - Implement ThreatModelEdit Save
The "save" button on the high-level threat model edit view is currently a no-op.

Implement the "save" functionality
Github Auth - Save to repository (implemented in backend)
Local Session - Do we download a copy?  We will need to figure out the best way to do this with the electron wrapper in the future.

**Additional context**
The graph functionality is still a WIP, this may break existing models.  v2 currently has a toast message to warn users that using this version may break existing models.
- Do we want to save a _copy_ as `<model_name>.v2.0.json` or similar? Or do we do the overwrite as it's always worked?

lreading:
**Summary**
Adds "save" functionality for browser.  This requires additional discussion, as saving vs downloading should _probably_ be different actions.  This does not add an implementation for electron, electron will prompt each time save is clicked.  Issue 383 was opened to address this.

Local sessions and git provider sessions can now save threat models
Closes #340


lreading:
I've created a discussion to flesh out the local session saving vs download a bit more: https://github.com/OWASP/threat-dragon/discussions/384

This can be handled as a separate issue once we have a consensus. :smile:

jgadsden (approved):
Definitely saving the file : 
'Demo Threat Model (1).json'
'Demo Threat Model (2).json'
'Demo Threat Model.json'

The saving of multiple copies can be worked on later

jgadsden:
rerunning the tests to see if the e2e test passes this time round

td.vue/src/i18n/en.js
```diff
@@ -66,6 +66,10 @@ const en = {
         diagrams: 'Diagrams',
         addNewDiagram: 'Add a new diagram...',
         threats: 'Threats',
+        errors: {
+            save: 'Error saving Threat Model.  Please check the developer console for more information'
```
ShubhamPalriwala, line 70: Very small nit, but there are 2-3 spaces after the full stop.

ShubhamPalriwala (approved):
Concept and Code Review Ack! Nice clean code with descriptive TODOs

Sorry for that small nit, I have an OCD!

td.vue/src/i18n/en.js
lreading: I wanted to double-check before commenting: Everything containing a period has a double-space at the end.  It's how I was taught... you know, back when dinosaurs were around.  I'll update all of them. :grin:

PR #392, opened 2022-03-29
update jquery

jgadsden:
**Summary**
Updates jquery

update jquery

**Other info**
@lreading I did not want to merge this directly, as I was not sure if the previously defined version of jquery was needed for something or not


lreading (approved):

jgadsden:
Thanks @lreading 
The zlib CVE is one that has caused all sorts of issues with other products. There is no prospect of development being done on zlib, at least in the short to medium term, so I have put it into the `.trivyignore ` file

PR #394, opened 2022-04-05
disable magnet to allow grab of entire element

jgadsden:
**Summary**
We found that the Actor, Process and Store diagram elements could only be grabbed using the element text, which can be fiddly for short text. If the user missed the grab then a dataflow is created, resulting in some confusion

disable magnet to allow grab of entire element

**Other info**
We have switched the magnets off for now. This means that data flows can only be created by dragging across from the palette. We need to fix this, so will raise an issue for that


jgadsden:
Note that we have failing tests because of the known node issue "pm ERR! Workspaces not supported for global packages"

PR #432, opened 2022-05-30
move to node 18 for github workflows

jgadsden:
**Summary**
There is [a known bug in npm](https://github.com/nodejs/node/issues/42397) versions 16.14.1 and 16.14.2. The github workflows were fixed at node version 16.14.0 until a fix was in place. Node 18.2.0 is now released and stable, so moving to that version instead

move to node 18 for github workflows

**Other info**
This reverts the change that temporarily fixed the node version to 16.14.0 in commit 82c9d49d2d3a315650a7aed7e1b3a3308a7b44a7

### DIFF
--- a/.github/workflows/browserstack.yaml
+++ b/.github/workflows/browserstack.yaml
@@ -13,10 +13,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


jgadsden:
Merging pull request after all checks have passed

PR #445, opened 2022-07-08
Update eslint and apply renaming

jgadsden:
**Summary**
Major update to eslint that required other updates as well. Not quite dependency hell, more like dependency purgatory :)

Eslint asks that components are more than one word, to avoid confusion in HTML, hence the renaming. All good now and linting and tests pass

Update eslint and apply renaming

**Other info**
I have had to disable some tests, I could not work out why they fail with the newer versions of Jest. I will raise an issue as a reminder to re-instate them:
```
    TypeError: Cannot read properties of undefined (reading 'shallow')

      31 |     [BRANCH_CLEAR]: (state) => clearState(state),
      32 |     [BRANCH_FETCH]: (state, branches) => {
    > 33 |         branches.forEach((branch, idx) => Vue.set(state.all, idx, branch));
         |                                               ^
      34 |     },
      35 |     [BRANCH_SELECTED]: (state, repo) => {
      36 |         state.selected = repo;

      at Function.set (node_modules/.pnpm/vue@2.7.3/node_modules/vue/dist/vue.runtime.common.dev.js:4368:17)
      at forEach (src/store/modules/branch.js:33:47)
          at Array.forEach (<anonymous>)
      at Object.BRANCH_FETCH (src/store/modules/branch.js:33:18)
      at Object.<anonymous> (tests/unit/store/modules/branch.spec.js:93:17)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
         "build:vue": "cd td.vue && npm run build",
         "clean": "npm-run-all clean:server",
         "clean:server": "cd td.server && npm run clean",
-        "codecov": "./node_modules/.bin/codecov",
+        "codecov": "echo 'Disabled for now: codecov'",
         "dev:vue": "cd td.vue && npm run dev",
         "dev:server": "cd td.server && npm run dev",
         "preinstall": "npx only-allow pnpm",
--- a/td.server/package.json
+++ b/td.server/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "build": "npm-run-all clean transpile",
     "clean": "rimraf dist",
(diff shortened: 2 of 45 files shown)


lreading:
I'm only able to look on my phone (on holiday, didn't bring a computer), but I'd venture a guess that the way jest is handling globals has changed. The error you pasted seems upset about the Vue object - iirc there was some special handling for that previously. Either way, great stuff!!!

lreading:
I also can't figure out how to formally approve from mobile, but I definitely approve! 😁

PR #475, opened 2022-08-07
revise logging to trace login request

jgadsden:
**Summary**
_**temporary pull-request**_

**Other info**


jgadsden:
The trivy scan failure has been fixed in pull request #474 so can be ignored here

jgadsden:
Merging this as logout will always have to be protected with authorisation

PR #478, opened 2022-08-09
no rate-limiting for test environments

jgadsden:
**Summary**
This reapplies the rate limiting if process.env.NODE_ENV 'development' is used, so rate limiting is only disabled if process.env.NODE_ENV === 'test'

no rate-limiting for test environments

**Other info**


jgadsden:
rate limiting can not be applied when in 'development' mode, as this is used by the end-to-end tests

Issue #468, opened 2022-08-05
multiple-line description support 


gassara-kys:
**Describe the bug**

If we have written a multi-line `High level system description`, it will be rendered on a single line in the model and report screens.

**Expected behaviour**

To be rendered in multiple lines in the above screens

**Environment**
 - Version: 2.0.0 (e.g. 1.3.1)
 - Platform: Web App (Web App / Desktop App)
 - OS: MacOS (MacOS / Windows / Linux)
 - Browser: chrome (e.g. chrome, firefox, safari)

**To Reproduce**

- create new model
- write description
- display dashboard & 

**Any additional context, screenshots, etc**

- submit mutiple-line description
  - ![image](https://user-images.githubusercontent.com/25426601/182982979-d1133484-eacc-49f7-a8ed-26793b047411.png)
- rendering single-line
  - ![image](https://user-images.githubusercontent.com/25426601/182983712-762588f1-3408-43ee-b422-0d3c6147e94f.png)

jgadsden:
Thanks @gassara-kys for finding this. We have multi-line descriptions in 1.x, so it should be relatively straight forward to reinstate this for version 2.0
I have marked it as a priority for version 2.0

Rets66:
There are some related issues.

I inputed the text to `Discription` and `Mitigations` like as follows:
- Discription
  <img width="1959" alt="スクリーンショット 2022-08-09 21 23 59" src="https://user-images.githubusercontent.com/17017861/183646762-0d512008-10fd-4bf0-b366-a23f89008c74.png">
- Mitigation
  <img width="787" alt="スクリーンショット 2022-08-09 21 24 22" src="https://user-images.githubusercontent.com/17017861/183646795-c055491f-19fa-4d66-9ce1-48481af2bee3.png">

But the out put were like this:
<img width="1959" alt="スクリーンショット 2022-08-09 21 22 37" src="https://user-images.githubusercontent.com/17017861/183647358-eb4cdfca-deb5-4153-8b07-593ef1e89c95.png">

PR #479, opened 2022-08-09
Fix some multi line issues

Rets66:
**Summary**

This fixes some multi line display issues like as follows:
1. When you review the report, `mitigations` inside of table don't support multi-line.
2. When you print the analyzed report out , `High level system description` and table under the diagram don't support it.

It's related with https://github.com/OWASP/threat-dragon/issues/468

I added CSS for the issue to fix.

**Other info**

### DIFF
--- a/td.vue/src/components/printed-report/ExecutiveSummary.vue
+++ b/td.vue/src/components/printed-report/ExecutiveSummary.vue
@@ -51,6 +51,7 @@
 .page {
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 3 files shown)


jgadsden (approved):
Many thanks @Rets66 for finding these and fixing them

jgadsden:
The failure in the build_docker_image stage is because this is the first commit by @Rets66 and so the username/password is withheld
This is not a problem with the changes, so OK to go ahead and merge

PR #480, opened 2022-08-10
always allow logout using GET

jgadsden:
**Summary**
Allow logout for unauthenticated / expired sessions

allow logout without authorisation

**Other info**
POST, PUT and DELETE will always require authentication
logout using GET will now succeed


jgadsden:
The failing e2e tests do not seem to be related to these changes, so going ahead with merge

PR #508, opened 2022-09-05
Add threat model json schema

jgadsden:
**Summary**
It would be good to provide a JSON schema that threat models could be validated against

add json schema for threat model

**Other info**


jgadsden:
Hello @pdxjohnny, here is the completed json schema. Is this looking OK? I have put it all in one 400 line file ... hope that is normal thing to do

docs/development/schema/owasp.threat-dragon.schema.json
```diff
@@ -0,0 +1,407 @@
...
+        "reviewer": {
+          "description": "The reviewer of the overall threat model",
+          "type": "string"
+        }
+      },
+      "required": [ "contributors", "diagrams" ]
+    }
+  },
+  "required": [ "summary", "detail" ]
+}
```
pdxjohnny, line 407: Will validate today

pdxjohnny (commented):
Thank you! Will run some code to confirm, but this looks great to me so far as I can see.

docs/development/schema/owasp.threat-dragon.schema.json
jgadsden: Note that this is the version 1.x schema, and we are hoping that version 2.0 Threat Dragon (which is still in development) will keep to this schema
At present a version 2.0 threat model will fail the schema - but we are working on that

jgadsden: and I have corrected an error where the 'tdVersion' was in the wrong place

jgadsden:
Merging this in to the main branch because more detail changes are being added from other branches

PR #605, opened 2023-02-14
Dashboard actions kept to equal height

### LINKED ISSUE #603: The three main options are not always the same height
**Describe the bug**
As the page width reduces the text flows onto a new line. As a result the boxes are not of equal height.

**Expected behaviour**
Each box should be the same height

**Environment**
 - Version: v2.0.0-beta
 - Platform: Web App
 - OS: MacOS
 - Browser: Firefox 109.0.1 (64-bit)

 **To Reproduce**
1. Visit https://www.threatdragon.com/#/dashboard
2. Adjust the width of the browser window until some text breaks onto a new line 

image

ChristopherHackett:
**Summary**
By applying height 100% we can get actions to have equal height. Remaining changes move the `2rem` from margin and places it as outer padding.

Dashboard actions kept to equal height. Closes #603

### DIFF
--- a/td.vue/src/components/DashboardAction.vue
+++ b/td.vue/src/components/DashboardAction.vue
@@ -1,6 +1,6 @@
 <template>    
     <b-col lg>
-        <b-jumbotron class="text-center">
+        <b-jumbotron class="text-center action-pane">
             <router-link :to="to">
                 <font-awesome-icon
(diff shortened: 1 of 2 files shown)


jgadsden (approved):
Thanks again @ChristopherHackett , nice to see these changes

PR #632, opened 2023-03-24
Fix #631 error while creation of threat model when using GitHub

### LINKED ISSUE #631: Unable to create a new threat model while using GitHub as provider
**Describe the bug**
When using GitHub as the provider, an error occurs when trying to create a new threat model.

**Expected behaviour**
When pressing Save button during the creation of a new threat model while using GitHub as provider, a new model.json file should be created on the GitHub repo.

**Environment**
 - Version: 2.0.1
 - Platform: Web App
 - OS: MacOS
 - Browser: Chrome

**To Reproduce**
- Clone the main branch of the repository
- Configure .env file for using GitHub as the provider
- Start the application
- Enter localhost:8080
- Login using GitHub
- Select the option to start a new model
- Select the repository and branch
- Try to save the new model, with or without changing some values of the new model, like name or description.
- A 500 error occurs during the put method to the update api

Screenshot 2023-03-24 at 11 44 56

Screenshot 2023-03-24 at 11 47 25

martinmarsicano:
**Summary**
This PR fixes #631

Allow the creation of a threat model while using GitHub as a provider by making the ThreatModelEdit call the create method of the server when on /create url.

**Other info**
- Creation of ThreatModelCreate (/crete) url for ThreatModelEdit component to differentiate when on creation from editing a threat model.
- Change THREATMODEL_CREATE action of the threatmodel store so the new threat model gets created.
- Creation of createAsync method on threatmodelApi to call /create api endpoint.
- Change from PUT to POST of /create api endpoint of server routes.

### DIFF
--- a/td.server/src/config/routes.config.js
+++ b/td.server/src/config/routes.config.js
@@ -42,7 +42,7 @@ const routes = (router) => {
     // removed because of security denial of service concerns (denial of models)
     //router.delete('/api/threatmodel/:organisation/:repo/:branch/:model', threatmodelController.deleteModel);
 
-    router.put('/api/threatmodel/:organisation/:repo/:branch/:model/create', threatmodelController.create);
+    router.post('/api/threatmodel/:organisation/:repo/:branch/:model/create', threatmodelController.create);
     router.put('/api/threatmodel/:organisation/:repo/:branch/:model/update', threatmodelController.update);
 };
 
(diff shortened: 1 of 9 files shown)


jgadsden (approved):
looks great, thanks @martinmarsicano . Also good to see some of the rough-and-ready code being cleaned up

jgadsden:
A couple of the unit tests are failing - are you OK with fixing these @martinmarsicano ?

martinmarsicano:
Now the test are running the same as last commit (except for Docker, because it fails doing login, I supposed that is something because it's running with my user or because of the branch).
Did you do the merge @jgadsden or do I?

PR #688, opened 2023-06-29
add repo select page query param

SilverC:
**Summary**:
Adds pagination to repository and branch selection screen. Currently only the first ~30 repositories a user has access to are shown with no way to view others. Fixes issue with repo URI being hardcoded to github.com on the same screens.

**Description for the changelog**:
Adds pagination to repository and branch selection screens.

**Other info**:
closes #547 

Thanks for submitting a pull request!
Please make sure you follow our code_of_conduct.md and our contributing guidelines contributing.md

PR #693, opened 2023-07-06
Add await to provider selection call

SilverC:
**Summary**:
Small edge case race condition occurs when user logs out and then without refreshing clicks the local login button resulting in the provider not being set yet. This is due to a new async call being made in the provider selection action which would allow the dashboard render to occur before the provider state mutation was committed. 

Add await to provider selection call to fix log out and network call cache edge case.

**Other info**:
Bug introduced in #688 


jgadsden (approved):
makes sense, thanks @SilverC , please go ahead and merge
I will fix up the Trivy Scan later, it is not applicable to us

PR #712, opened 2023-08-04
Update jwt.helper.js to increase Github JWT Token duration

HugoReeves:
**Summary**:
Currently Github users are logged out after 5 minutes. This makes it impossible to save changes without refreshing and losing all unsaved changes. As far as I can tell, refresh tokens have not been implemented currently. Increasing the JWT token duration to 1 day should make it more practical to edit Github models. A longer term solution is to finish implementation of refresh tokens.

Update access token and refresh token duration.
Access token updated to 1 day from 5 minutes.
Refresh token updated to 7 days from 24 hours.

**Other info**:
None


jgadsden (approved):
thanks @HugoReeves , these changes make good sense and I agree with the values
I had not come across it earlier because I tend to use local login for testing
did you want to raise an issue for the refresh token?

jgadsden:
there is a failing unit test because it is searching for '5m', did you want to fix that up? otherwise I can modify this

HugoReeves:
@jgadsden The unit test should be updated now.

PR #731, opened 2023-08-29
Initial finnish translation.

mikkolehtisalo:
This pull request contains an Initial Finnish language translation for the user interface.  

After translating the strings I checked how they work on the UI, and did some fine tuning. I did not translate STRIDE etc terms, because I don't think they should be translated. Working with them translated would be probably awkward.

### DIFF
--- a/td.vue/src/desktop/menu.js
+++ b/td.vue/src/desktop/menu.js
@@ -15,6 +15,7 @@ var mainWindow;
 import deu from '@/i18n/de.js';
 import ell from '@/i18n/el.js';
 import eng from '@/i18n/en.js';
+import fin from '@/i18n/fi.js';
 import fra from '@/i18n/fr.js';
 import hin from '@/i18n/hi.js';
 import por from '@/i18n/pt.js';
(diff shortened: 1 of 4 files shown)


jgadsden (changes requested):
Hello @mikkolehtisalo , very good to see this translation file, thanks.

There are a couple of other files that need changes for the desktop : `td.vue/src/desktop/menu.js` and `td.vue/tests/unit/desktop/menu.spec.js`
are you OK with making these changes?

jgadsden:
Agreed that terms such as STRIDE, LINDDUN, CIA etc are best left untranslated.

Do not worry about the PR pipeline / Build docker image test failure - that is due to permissions not being high enough

mikkolehtisalo:
That should do it?

jgadsden (approved):
superb @mikkolehtisalo , that is perfect thank you
do you want to go ahead and merge?

jgadsden:
merging this to main, thanks again @mikkolehtisalo

PR #779, opened 2023-11-07
Fix searchAsync to *actually* use the access_token (and pagination) correctly

quine:
**Summary**:
With `GITHUB_USE_SEARCH` was enabled, calling `searchAsync` threw an error, as the `page` parameter was being passed in first (but wasn't used in the function).

- Fix `searchAsync` to enable repo search to work (and use pagination) correctly


jgadsden (approved):
thanks for finding this @quine , all good with the params now and ready to merge

jgadsden:
going ahead with the merge - the Docker action failing is the usual problem and not to do with this PR

Issue #426, opened 2022-05-13
Support storage backends other than github


danielpops:
**Describe what problem your feature request solves**
Instead of using github repositories for storage of threat models, I'd like to store them all in an alternative centralized storage mechanism (I have a strong affinity to AWS, so s3 would be my preference)

**Describe the solution you'd like**
Introduce a storage mechanism configuration and support options like which cloud provider (aws, azure, google) and any relevant configuration options for each one (for AWS, this would probably just be bucket name and region). Of course, this introduces a new dependency on IAM credentials for the service, which I'd say is an exercise left to the reader (I'd personally launch threat-dragon as a containerized workload in AWS and attach an IAM role, but others assumedly would want to set AWS_ACCESS_KEY_ID etc.)

jgadsden:
Certainly sounds good to me - are you able to contribute time for this @danielpops (no harm in asking I hope)?

steve-winter:
Implementation should be very similar to #1 so I will progress a PR. Highlighting now so you can shut this down if you dont want it. 

Intent is:
1. New auth provider as 'AWS', which triggers the AWS OIDC - https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/sso-oidc/
2. Once auth'd as AWS create a new repository and use the S3 client - https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/

May need to 'default' the branch screen to show a generic branch equivalent, possibly use S3 versioning.

jgadsden:
excellent that you can work on this @steve-winter , thanks for taking this one on as well

Issue #61, opened 2021-04-23
Support Azure OAuth


lreading:
**Describe what problem your feature request solves**
Add ability to use multiple passport connectors, such as Azure AD.

**Describe the solution you'd like**
I'd like to add the Azure AD Passport connector for authentication/authorization.  I _think_ it would make the most sense to have this be enabled via configuration, and also allow for multiple passport strategies.  For example: `export PASSPORT_STRATEGIES="['github', 'azuread-openidconnect']"` or `export PASSPORT_STRATEGIES="['github']"`
This would open up the possibility of adding other passport strategies/connectors in the future.

**Additional context**
Some questions I have:
- Because Azure AD requires a bunch more config options, would it make sense to implement something like [dotenv](https://github.com/motdotla/dotenv) to make configuration a bit easier?
  - If so, should that be a separate feature / PR to keep the scope somewhat limited?
- Does it make sense to restructure how the passport authentication is configured to make it easier to add other authentication types in the future?

I'm eager to add this functionality, but wanted to see if it's a desired feature, and what the appetite is as far as the size and scope of pull requests.

jgadsden:
Hello @lreading , yes, definitely an appetite for what you are describing. We are looking towards Threat  Dragon 2.0 this year and expanding the ways to authenticate is central to this.
We have a [roadmap](https://owasp.org/www-project-threat-dragon/) for version 2.0, and we would like various ways to authenticate - sort of similar to https://app.diagrams.net, as we are also wanting to use mxgraph.

Certainly will accept large pull requests, but try and break them up if you can, and the scope is the whole project - do not hold yourself back

Cheers, Jon

jgadsden:
@lreading I would normally assign this issue to you, but you have to be a contributor - so add some code and then we can get you on the contributors list. I have assigned it to myself for now

lreading:
I somehow missed the multi auth when I looked at the roadmap. :facepalm:  

A few more questions:
- For Azure, specifically, is there a preference for back-end storage of models, such as azure table storage?
- Would you prefer this to target the main trunk or the 2.0 branch?
- Do you want feature requests / PRs for individual items on the 2.0 roadmap, or do you plan to use a github project (I see one for 1.5)?
- If targeting 2.0, would you like help refactoring to ES6 before adding additional code? 

Thanks again!

jgadsden:
[...]

PR #801, opened 2023-11-25
Server config driven providers on HomePage.vue

steve-winter:
**Summary**:
As a pre-req for #426 and #61 enabling the homepage to only render login options that are configured. Ignored for Electron. 

Select which login options to render based on the configured environment variables, currently this is a check if the related **_CLIENT_ID is configured. 

**Other info**:
Environment variables likely require refactoring.


jgadsden (approved):
Good to see this @steve-winter , much appreciated, please go ahead and merge

PR #836, opened 2024-01-12
Navbar fixed

### LINKED ISSUE #835: Navbar Elements Not Visible on Mobile Screen When Toggled
**Describe the bug**:
When viewing the website on a mobile screen and toggling the navbar, the navbar elements do not appear as expected. It seems that the navbar is overlapping with the background, causing the elements to be hidden. This issue makes it difficult for users to access the navigation menu and navigate through the website on smaller screens. It requires investigation and resolution to ensure a seamless user experience across different devices.

**Expected behaviour**:
The elements should be visible and accessible by the viewers
**Environment**:

- Version: (e.g. 1.3.1)
- Platform: (Web App / Desktop App)
- OS: (MacOS / Windows / Linux)
- Browser: (e.g. chrome, firefox, safari)

**To Reproduce**:
 open the website 
webpage

It should look like :
solution

SharanRP:
**Summary**:
Fixed the navbar for mobile screens

Changes made in the toggle button 

**Other info**:
closes #835


example.env
```diff
@@ -4,11 +4,11 @@ BITBUCKET_SCOPE=repository:write
 BITBUCKET_WORKSPACE=workspace_name
 
 ENCRYPTION_KEYS='[{"isPrimary": true, "id": 0, "value": "11223344556677889900aabbccddeeff"}]'
-ENCRYPTION_JWT_SIGNING_KEY=asdfasdfasdf
-ENCRYPTION_JWT_REFRESH_SIGNING_KEY=fljasdlfkjadf
+ENCRYPTION_JWT_SIGNING_KEY=0c43457e9106e54da10f17f6d3bf9def
```
jgadsden, line 7: we require that real keys are not used here, as this is only an example file

td.vue/src/components/Navbar.vue
jgadsden, line 1: This is the real change, if I understand correctly, so if the other files could be reverted back to their unmodified state that would be great

jgadsden (changes requested):

SharanRP:
@jgadsden I have made the required changes

jgadsden (approved):
good to have this, thanks @SharanRP and will go ahead and merge

Issue #844, opened 2024-01-23
Multiselect is not working


Sondrebf:
**Describe the bug**:

Multiselect wont let me drag all selected components

**Expected behaviour**:

Before i could select all the components i wanted to move and move them. Now only one component is moved-

**Environment**:

- Version: 2.1.3-latest
- Browser: chrome -latest

jgadsden:
Hello @Sondrebf , can you provide the version that it worked OK for?

Sondrebf:
Sorry i dont remember, i used the web version last year and it worked.

raghav1030:
I have somehow solved the issue. Hey @Sondrebf, can you please confirm if you're talking about this behavior?

https://github.com/OWASP/threat-dragon/assets/94848646/4423fae9-ed95-432d-95d5-74bd4e3fac7f

Sondrebf:
Yes, thats what i wanted, and had before 👍 Good!

jgadsden:
yes, many thanks @raghav1030, please go ahead

PR #849, opened 2024-01-30
enable multi selection of components

raghav1030:
**Summary**:

This PR enables multi-selection of components and resolves issue #844.


jgadsden:
We need to check that there are no side effects with this, as this code was there to correctly select the properties of the diagram component

jgadsden (approved):
Tested and I agree that my concerns about the selection of attributes was unfounded, and all tests out correctly. Thanks for this @raghav1030 and please go ahead and merge

Issue #846, opened 2024-01-25
No prompt to save on exit


antnks:
**Describe the bug**:
There is no prompt to save changes when quitting the app.

**Expected behaviour**:
If there were changes made, the app should prompt to save or discard changes to prevent unexpected data loss.

**Environment**:

- Version: 2.1.3
- Platform: Desktop App
- OS: MacOS arm64

**To Reproduce**:
1. Open the app and create a new model
2. Add diagram
3. Draw some diagram elements
4. Hit Cmd+Q to exit app
5. App exits, all unsaved changes are lost

raghav1030:
Hey @jgadsden! I've been working on implementing this feature to prompt the user to save the model before quitting the application. I've made some progress, but I'm encountering an issue. When the user closes the model by clicking on the close button in Windows, the application prompts the user to verify if they want to save the model or not.

However, the problem arises when the user doesn't close the application or the model, but instead clicks on the threat model icon in the navbar or any other button (like logout) that sends them to a different route. In such cases, the route changes but the prompt doesn't occur.

I've tried to implement the same logic when the graph component gets destroyed, but the function is asynchronous, and the destroyed or beforeDestroy lifecycle functions don't support the asynchronicity of the functions called within them. So, I'm struggling to find a way to prompt the user to save the model when the routes change.

You can find a discussion about a similar issue in https://github.com/vuejs/vue/issues/7209.

Here is the video depicting both the behavior : 

a) When the user simply closes the whole application :-

https://github.com/OWASP/threat-dragon/assets/94848646/60f365f3-cb11-46e6-922f-08e0f3f41153

b) When the user is redirected to any other route :-

https://github.com/OWASP/threat-dragon/assets/94848646/9afe5223-0e48-434e-bdc0-cc1a9df3395c

jgadsden:
thanks for working on this @raghav1030 
there are two issues here:
* we need a guard on the route to #/dashboard, for both desktop and webapp
* also in desktop we need to block and request to close the model when the electron app is about to close, from the menu  item 'Quit Threat Dragon' and also when using the window close button

raghav1030:
> thanks for working on this @raghav1030 there are two issues here:
> 
> * we need a guard on the route to #/dashboard, for both desktop and webapp
> * also in desktop we need to block and request to close the model when the electron app is about to close, from the menu  item 'Quit Threat Dragon' and also when using the window close button

[...]

PR #852, opened 2024-02-04
Prompt user to save model before quitting the app

raghav1030:
**Summary**:

Prompts the user to save the model before quitting. Solves #846 

Controlled the way Electron was closing the app previously. However, the guard on the route to #/dashboard has not been implemented yet. We can address this in a separate issue to better organize the code.

### DIFF
--- a/td.vue/public/preload.js
+++ b/td.vue/public/preload.js
@@ -6,6 +6,7 @@ if (process.env.IS_TEST === 'true') {
 
 contextBridge.exposeInMainWorld('electronAPI', {
     // renderer to electron main
+    appClose: () => ipcRenderer.send('close-app'),
     updateMenu: (locale) => ipcRenderer.send('update-menu', locale),
     modelClosed: (fileName) => ipcRenderer.send('model-closed', fileName),
     modelModified: (modified) => ipcRenderer.send('model-modified', modified),
@@ -15,6 +16,7 @@ contextBridge.exposeInMainWorld('electronAPI', {
     modelSave: (modelData, fileName) => ipcRenderer.send('model-save', modelData, fileName),
 
     // electron main to renderer
+    onCloseAppRequest: (callback) => ipcRenderer.on('close-app-request', callback),
     onCloseModelRequest: (callback) => ipcRenderer.on('close-model-request', callback),
     onNewModelRequest: (callback) => ipcRenderer.on('new-model-request', callback),
     onOpenModel: (callback) => ipcRenderer.on('open-model', callback),
--- a/td.vue/src/components/Graph.vue
+++ b/td.vue/src/components/Graph.vue
@@ -111,7 +111,8 @@ export default {
         }
     },
     destroyed() {
-        diagramService.dispose(this.graph);
+      diagramService.dispose(this.graph);
+      this.$store.dispatch(tmActions.notModified);
     }
 };
 </script>
--- a/td.vue/src/desktop/desktop.js
+++ b/td.vue/src/desktop/desktop.js
@@ -56,34 +56,12 @@ async function createWindow () {
         menu.setMainWindow(mainWindow);
     });
 
-    mainWindow.on("close" , (event) => {
-        const modified = menu.model.isModified
-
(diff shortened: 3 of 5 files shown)


jgadsden:
Hello @raghav1030 , there might be a better way of handling the application close event
In a way this is similar to the Close Model menu action - we want to close the model so we request the renderer to do this, then the renderer notifies electron that it is safe to do so
So I think we should use the existing requests already in place, which allows for a consistent 'save changes?' dialog, and then allow the renderer to notify electron when it is safe to close the application

raghav1030:
> Hello @raghav1030 , there might be a better way of handling the application close event In a way this is similar to the Close Model menu action - we want to close the model so we request the renderer to do this, then the renderer notifies electron that it is safe to do so So I think we should use the existing requests already in place, which allows for a consistent 'save changes?' dialog, and then allow the renderer to notify electron when it is safe to close the application

Hey @jgadsden, could you please verify if you're referring to this exact same behavior?

https://github.com/OWASP/threat-dragon/assets/94848646/1eb42e44-c707-4f22-beee-610578fdf0b9

jgadsden (changes requested):
I think we should use the existing IPC between electron and the renderer
This allows the renderer to be the authority on whether to exit the application or not, and also allows reuse of the existing renderer dialog 
@raghav1030 could you consider this PoC :

```
diff --git a/td.vue/public/preload.js b/td.vue/public/preload.js
index 6300f36b..f0454e18 100644
--- a/td.vue/public/preload.js
+++ b/td.vue/public/preload.js
@@ -8,2 +8,3 @@ contextBridge.exposeInMainWorld('electronAPI', {
     // renderer to electron main
+    appClose: () => ipcRenderer.send('close-app'),
     updateMenu: (locale) => ipcRenderer.send('update-menu', locale),
@@ -17,2 +18,3 @@ contextBridge.exposeInMainWorld('electronAPI', {
     // electron main to renderer
+    onCloseAppRequest: (callback) => ipcRenderer.on('close-app-request', callback),
     onCloseModelRequest: (callback) => ipcRenderer.on('close-model-request', callback),
diff --git a/td.vue/src/desktop/desktop.js b/td.vue/src/desktop/desktop.js
index 061199a6..ec34dfef 100644
--- a/td.vue/src/desktop/desktop.js
+++ b/td.vue/src/desktop/desktop.js
@@ -22,2 +22,4 @@ protocol.registerSchemesAsPrivileged([
 let loadingScreen;
+let runApp = true;
+
 async function createWindow () {
@@ -58,2 +60,9 @@ async function createWindow () {
 
+    mainWindow.on("close" , (event) => {
+        if (runApp) {
+            event.preventDefault();
+            mainWindow.webContents.send('close-app-request');
+        }
+    });
+
     if (electronURL) {
@@ -110,2 +119,3 @@ app.on('ready', async () => {
 
+    ipcMain.on('close-app', handleCloseApp);
     ipcMain.on('update-menu', handleUpdateMenu);
@@ -132,2 +142,8 @@ app.on('open-file', function(event, path) {
 
+function handleCloseApp (_event) {
+    logger.log.debug('Close application request from renderer ');
+    runApp = false;
+    app.quit();
+}
+
 function handleUpdateMenu (_event, locale) {
diff --git a/td.vue/src/main.desktop.js b/td.vue/src/main.desktop.js
index b3612a24..ddb4e3d8 100644
--- a/td.vue/src/main.desktop.js
+++ b/td.vue/src/main.desktop.js
@@ -31,2 +31,12 @@ const getConfirmModal = () => {
 
+// request from electron to renderer to close the application
+window.electronAPI.onCloseAppRequest(async (_event) =>  {
+    console.debug('Close application request');
+    if (!app.$store.getters.modelChanged || await getConfirmModal()) {
+        console.debug('Closing application');
+        // send request back to electron server to close the application
+        window.electronAPI.appClose();
+    };
+});
+
 // request from electron to renderer to close the model
```

jgadsden (approved):
thanks @raghav1030 , going ahead with merge

Issue #892, opened 2024-03-01
More than one threat of same type can be added for same element of chosen threat model 


itsdheerajdp:
**Describe the bug**:
Currently we can add same type of threat infinite number of times for same element of threat model.
Ex. For STRIDE Model , We can add Spoofing type of threat 'x' Number of times which is not desirable for threat modeling.

[screen-capture.webm](https://github.com/OWASP/threat-dragon/assets/107138786/9fda712e-725d-4b55-bbeb-402206509ba9)

**Expected behaviour**:
For Same threat model , One element can have at most of one threat of one type 
ex. For a element in STRIDE model , the element can have at most of one spoofing type threat.

**Environment**:

- Version: (e.g. 1.3.1)
- Platform: (Web App / Desktop App)
- OS: (MacOS / Windows / Linux)
- Browser: (e.g. chrome, firefox, safari)

**To Reproduce**:

**Any additional context, screenshots, etc**:

itsdheerajdp:
@jgadsden If you found this as a bug then i'll work on this ...

jgadsden:
This is not a bug - it is very valid, and is often the case, that more than one threat of the same type is identified for the same component
the application is working as expected and closing this issue as "Won't Do"

PR #893, opened 2024-03-01
Fix: Adding Same type of Threats

itsdheerajdp:
**Summary**:
Fixes:[#892 ]

Now We can't add same type of threat to same element of chosen threat model . Now our webpage will behave like below on adding same type of threat more than once.

[screen-capture (1).webm](https://github.com/OWASP/threat-dragon/assets/107138786/f2cd178e-bd10-443b-a207-8a8e2d5b21ff)

**Other info**:

### DIFF
--- a/td.vue/src/components/ThreatEditDialog.vue
+++ b/td.vue/src/components/ThreatEditDialog.vue
@@ -240,7 +240,12 @@ export default {
         },
         updateThreat() {
             const threatRef = this.threat;
-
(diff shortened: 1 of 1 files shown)


jgadsden (changes requested):
This is not a bug - it is very valid, and is often the case, that more than one threat of the same type is identified for the same component
the application is working as expected and closing this pull-request as "Won't Do"

Issue #894, opened 2024-03-01
New Threats are getting added when close the add new threats window is closed


itsdheerajdp:
**Describe the bug**:
Currently if we are clicking on [New Threat] Button then it will add the threat however if click on apply or not. By clicking on close button is also resulting in adding new threats. Visuals are shown below

[screen-capture.webm](https://github.com/OWASP/threat-dragon/assets/107138786/ae5936bb-e95d-4dfc-94f9-2974dc88f5a8)

**Expected behaviour**:
If we click on close button then new threat should not be added. New threat should only be added on clicking the [Apply] Button.

**Environment**:

- Version: (e.g. 1.3.1)
- Platform: (Web App / Desktop App)
- OS: (MacOS / Windows / Linux)
- Browser: (e.g. chrome, firefox, safari)

**To Reproduce**:

**Any additional context, screenshots, etc**:

jgadsden:
this is not a bug, in that this is the intended behavior
At  present the threat is removed by using the 'Remove' button, in a similar way to adding a new diagram, and not on closing the window
if we change this then we are breaking the consistency of the application

itsdheerajdp:
@jgadsden i thought if a user accidently clicked the New Threat Button but don't want to add the new threat so he close the window but still the new threat created which will cause a extra effort for him to remove the threat.

jgadsden:
I have created a list of actions for the threat creation and modification in #899 , so closing this issue

PR #895, opened 2024-03-01
Fixes: Add New Threat Window

itsdheerajdp:
**Summary**:
Now when we click on close window icon or we just close the window of adding new threat then new threat will not be added. New threat will only be added if click on Apply Button . It will solve the inconsistency in the new threat adding feature. And now our app will behave like this.

[screen-capture.webm](https://github.com/OWASP/threat-dragon/assets/107138786/7df126a8-5fd1-46b8-8998-0109896bcca7)

Add New Threat Window 

**Other info**:

Fixes : #894

### DIFF
--- a/td.vue/src/components/ThreatEditDialog.vue
+++ b/td.vue/src/components/ThreatEditDialog.vue
@@ -134,7 +134,7 @@
                     class="float-left"
                     @click="immediateDelete()"
                 >
-                    {{ $t('forms.remove') }}
+                    {{ $t('forms.remove') }} 
                 </b-button>
                  <b-button
                     variant="secondary"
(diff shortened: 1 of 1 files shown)


jgadsden (changes requested):
At  present the threat is removed by using the 'Remove' button, in a similar way to adding a new diagram, and not on closing the window
if we change this then we are breaking the consistency of the application

itsdheerajdp:
@jgadsden  i thought if a user accidently clicked the New Threat Button but don't want to add the new threat so he close the window but still the new threat created , which will cause a extra effort for him to remove the threat.

jgadsden:
@itsdheerajdp I have created an issue where the actions applied to the threat generation and modification are listed: #899 
we will need to follow that , so am marking this pull request as "Won't Do"

Issue #902, opened 2024-03-05
Provide translation for Demo Model Page


dheerajsingh89:
**Describe what problem your feature request solves**:
Currently , In the [Select Demo Page] User can't access the Demo models name in any language other than English as shown below

**Describe the solution you'd like**:
For Better User Experience the Demo models should also have used i18n 

**Additional context**:

![Screenshot 2024-03-05 192651](https://github.com/OWASP/threat-dragon/assets/107138786/7a465bb4-f2a0-4a26-bdd1-d637d28137b2)

jgadsden:
thanks @itsdheerajdp , this would be a valuable addition, assigning to you if you can also implement it

itsdheerajdp:
@jgadsden  I'll do this

itsdheerajdp:
@jgadsden I have come up with a solution. It'll be great if i got your suggestion on this.

jgadsden:
closing this as "won't do", it is better that the model names themselves are kept in English because the model itself is in English

PR #912, opened 2024-03-12
translation for Demo Model Page

itsdheerajdp:
**Summary**:

Now Our [Select Demo Model] Page is integrated with i18n and its content is available for other languages also as shown below.

Screenshot 2024-03-12 154903

Screenshot 2024-03-12 154913

Screenshot 2024-03-12 154936

Screenshot 2024-03-12 161742

translation for Demo Model Page

**Other info**:
Fixes : #902


itsdheerajdp:
@jgadsden what do  you say about this?

itsdheerajdp:
@jgadsden can i get you review on this ?

td.vue/src/i18n/ar.js
```diff
@@ -47,7 +47,12 @@ const ara = {
         }
     },
     demo: {
-        select: 'حدد نموذج تهديد تجريبي من القائمة أدناه'
+        select: 'حدد نموذج تهديد تجريبي من القائمة أدناه',
+        "New Threat Model": "نموذج تهديد جديد",
```
jgadsden, line 51: the names are not ideal, such as "New Treat Model", and should be renamed to something more in keeping with the rest of the attribute naming

td.vue/src/i18n/ar.js
```diff
@@ -47,7 +47,12 @@ const ara = {
...
+        "Demo Threat Model": "نموذج تهديد تجريبي",
+        "Version 2 New Model": "الإصدار ٢ نموذج جديد",
+        "Version 2 Demo Model": "الإصدار ٢ نموذج تجريبي"
```
jgadsden, line 54: I am unsure that the translations are correct - could you provide translations only for those languages that you speak natively? Translation programs tend to struggle over technical terms like this

td.vue/src/i18n/ar.js
jgadsden, line 51: the names are not ideal, such as "New Treat Model", and should be renamed to something more in keeping with the rest of the attribute naming

jgadsden (changes requested):

td.vue/src/i18n/ar.js
itsdheerajdp: > the names are not ideal, such as "New Threat Model", and should be renamed to something more in keeping with the rest of the attribute naming

Actually I gave such names because , these names i am getting in the front end page (SelectDemoModel.vue) as {{model.name}} from the backend 
if you want i can change through the backend or you can suggest any other better solution ??

jgadsden:
@itsdheerajdp could you consider this patch that will provide consistent attribute names and translate the strings only if we are fluent in the language (otherwise it could go wrong with Google translate):

[pr912.patch](https://github.com/OWASP/threat-dragon/files/14969700/pr912.patch)

itsdheerajdp:
> @itsdheerajdp could you consider this patch that will provide consistent attribute names and translate the strings only if we are fluent in the language (otherwise it could go wrong with Google translate):
> 
> [pr912.patch](https://github.com/OWASP/threat-dragon/files/14969700/pr912.patch)

Sure I'll do it

PR #914, opened 2024-03-12
Also build for arm64

jksolbakken:
**Summary**:
Builds Docker images also for ARM. Improves developer experience on newer Macs with ARM CPU (Mx)
Adds necessary changes to the affected GitHub workflows

add build for arm64 docker image in GitHub workflows

**Other info**:
The extra build time should be bearable, if not the builds for each arch can be run in parallel using a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs).

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -186,6 +186,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4.1.1
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v3.1.0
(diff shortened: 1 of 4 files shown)


jgadsden (approved):
superb, many thanks @jksolbakken

jksolbakken:
The build fails because of missing secrets for login to Docker Hub. Maybe remove the login step and do not push from the PR workflow?

jksolbakken:
I will update this PR to run the build step before the login for PRs. That way we can determine if the image build itself works without the need for elevated privs. The other steps should work as before when you merge. Secrets are not available if the workflow was caused by an external PR.

jgadsden (approved):
thanks agin @jksolbakken , and will go ahead and merge

PR #942, opened 2024-04-11
Stop report options moving and obscuring language drop-down

michaelgrahamevans:
**Summary**:

When on the report page, scrolling down the content causes the report options bar to move upwards slightly. Additionally, the report options bar appears above the language selection drop-down menu in the nav bar.

This change tweaks the CSS to correct the heights and margins of the report options bar so that it does not move and it changes the z-index of the report options bar to place it below the language drop-down.

**Other info**:

Before:

report-options-bar-before

language-drop-down-before

After:

report-options-bar-after

language-drop-down-after

### DIFF
--- a/td.vue/src/views/ReportModel.vue
+++ b/td.vue/src/views/ReportModel.vue
@@ -156,7 +156,7 @@
     width: 100%;
     background-color: $white;
     padding-top: 15px;
-    z-index: 9999;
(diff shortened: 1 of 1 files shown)


jgadsden (approved):
a good improvement, thanks @michaelgrahamevans

PR #951, opened 2024-04-18
Fix typos in demo threat descriptions

alexvorndran:
**Summary**:

This pull request fixes some minor typos in the threat descriptions of the demo models (and a test fixture). These aren't critical in the sense that they'd distort the meaning, but the fix may help to improve the first impression of Threat Dragon.

Fix typos in demo threat descriptions

**Other info**:

Based on my understanding of the end-to-end test that uses the fixture in question ([`td.vue/tests/e2e/specs/threatmodel/import.cy.js`](https://github.com/OWASP/threat-dragon/blob/01b1bcc35365bb376e81aba3a480f2789a122660/td.vue/tests/e2e/specs/threatmodel/import.cy.js#L83)), the change shouldn't have any effect on the test because the changed description isn't checked. The test is also skipped at the moment.

### DIFF
--- a/ThreatDragonModels/demo-threat-model.json
+++ b/ThreatDragonModels/demo-threat-model.json
@@ -88,7 +88,7 @@
                 {
                   "status": "Open",
                   "severity": "Medium",
-                  "description": "An attacker could obtain the DB credentials ans use them to make unauthorised queries.",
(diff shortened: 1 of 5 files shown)


jgadsden (approved):
Thanks for this @alexvorndran , good to fix typos and going ahead with the merge

PR #952, opened 2024-04-18
Fix undo for edges

### LINKED ISSUE #428: Undo not fully functional
**Describe the bug**
While the undo operation works for deleted processes, actors, and data storages, it doesn´t work for arrows (data flows). If you delete one of the aforementioned diagrams which were previously connected with arrows, after "undo", the diagrams will show up again, but the arrows are gone, together with the threats defined for them. 
Undo also doesn´t work if you just delete an arrow and want to undo it. 

**Expected behaviour**
After accidental deletion of an arrow, or diagrams connected with arrows, the undo button should make the arrows reappear. It is particularly important when threats were already defined for those data flows.

**Environment**
 - Version: 2.0
 - Platform: Web App Beta
 - OS: Windows
 - Browser: Chrome

**To Reproduce**
1. Create two actors.
2. Connect them with a data flow arrow.
3. Delete one of the actors. (The arrow should also disappear.)
4. Undo the operation. (The actor reappears but the connection arrow doesn´t.)

**Further information**
- Undo button does not undo deletion of objects. This is a big deal because objects are too easy to remove accidentally.
- If it is too hard to undo deletion, perhaps an “are you really sure” button would help with accidental deletes?

moerketh:
**Summary**:
This change should fix the undo of edges in the model. Both when deleting a node with connected edges and a single edge.

Here is the undo behavior before this change (on the public website):

https://github.com/OWASP/threat-dragon/assets/1309462/20b4f66a-b992-4454-873c-48c83c1af27f

Here is the undo behavior after this change (locally hosted):

https://github.com/OWASP/threat-dragon/assets/1309462/033493f5-3991-4715-989b-e07df08145b9

Register Edges in the x6 model instead of registering Edges as Nodes.

**Other info**:
Closes #428

Line 14 might need a similar change, but I am not familiar with the compatibility concerns.

### DIFF
--- a/td.vue/src/service/x6/shapes/index.js
+++ b/td.vue/src/service/x6/shapes/index.js
@@ -14,12 +14,12 @@ import { TrustBoundaryCurveStencil } from './trust-boundary-curve-stencil.js';
 Graph.registerNode('trust-broundary-curve', TrustBoundaryCurve);
 
 Graph.registerNode('actor', ActorShape);
-Graph.registerNode('flow', Flow);
+Graph.registerEdge('flow', Flow);
 Graph.registerNode('process', ProcessShape);
 Graph.registerNode('store', StoreShape);
 Graph.registerNode('td-text-block', TextBlock);
 Graph.registerNode('trust-boundary-box', TrustBoundaryBox);
-Graph.registerNode('trust-boundary-curve', TrustBoundaryCurve);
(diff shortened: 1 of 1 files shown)


jgadsden (approved):
many thanks @moerketh , good to such an elegant bug fix

PR #959, opened 2024-05-15
Set model storage path via env var

jason-meredith-rewind:
**Summary**:
As per the documentation found here: https://owasp.org/www-project-threat-dragon/docs-2/getting-started#opening-an-existing-model

> Note that Threat Dragon is fairly strict on where the threat models can be stored. The threat models must be under a parent directory called ‘ThreatDragonModels’ and the JSON file must then be stored in a sub-directory with the same name as the model.
> As an example, shown is a directory containing two models ‘test-reports’ and ‘New Threat Model’ under the directory ‘ThreatDragonModels’. This directory structure has been carried over from Threat Dragon versions 1.x, and in future it may become less strict.

I was hoping there would be a way of configuring where saved models are stored on a repo (ie in a `/doc` directory), and seeing this thought to add the functionality myself.

With this change, if no changes are made to the environment variables nothing changes in the directory structure - it is entirely opt-in and backwards compatible. If you would like to customize where models are stored, you can now set the `REPO_ROOT_DIRECTORY` environment variable. You can also override it per platform, ie `GITHUB_REPO_ROOT_DIRECTORY`, `BITBUCKET_REPO_ROOT_DIRECTORY`, etc.

Customizable model storage path

### DIFF
--- a/td.server/src/env/Bitbucket.js
+++ b/td.server/src/env/Bitbucket.js
@@ -20,7 +20,8 @@ class BitbucketEnv extends Env {
             { key: 'ENTERPRISE_PROTOCOL', required: false, defaultValue: 'https' },
             { key: 'USE_SEARCH', required: false, defaultValue: false },
             { key: 'SEARCH_QUERY', required: false },
-            { key: 'WORKSPACE', required: false }
+            { key: 'WORKSPACE', required: false },
+            { key: 'REPO_ROOT_DIRECTORY', required: false }
         ];
     }
 }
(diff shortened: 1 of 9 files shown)


jgadsden:
many thanks for this @jason-meredith-rewind and ignore one of the broken tests (Desktop e2e) as this is a know issue
Could you have a look at the server unit tests? These should run OK so I guess the changes need some fixups in the unit tests
You can run the server unit tests by themselves with `npm run test:server`

jgadsden (approved):
many thanks for this @jason-meredith-rewind 
all good to merge

PR #996, opened 2024-07-03
Aligned scope param default value with oauth spec

### LINKED ISSUE #995: Malformed gitlab scope default value
**Describe the bug**:
The gitlab integration is fuzzy.
E.g. the GITLAB_SCOPE env var defaults to **_'read_user,read_repository,write_repository,profile'_**.
The resaults in the following gitlab error: **_The requested scope is invalid, unknown, or malformed._**

**Expected behaviour**:
The correct default value for the GITLAB_REDIRECT_URI would be **_'read_user read_repository write_repository profile'_**.
So separation by spaces instead of commas.

This is also defined [here](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3):

> The value of the scope parameter is expressed as a list of space-delimited, case-sensitive strings.

**Environment**:

- Version: 2.2.0
- Platform: Docker

**To Reproduce**:

ch-lepp:
**Summary**:
Closes #995 

The scope params default value is separated by spaces instead of commas as defined by [the oauth spec](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3).

### DIFF
--- a/td.server/src/providers/gitlab.js
+++ b/td.server/src/providers/gitlab.js
@@ -28,7 +28,7 @@ const getGitlabUrl = () => {
  * @returns {String}
  */
 const getOauthRedirectUrl = () => {
-    const scope = env.get().config.GITLAB_SCOPE || 'read_user,read_repository,write_repository,profile';
+    const scope = env.get().config.GITLAB_SCOPE || 'read_user read_repository write_repository profile';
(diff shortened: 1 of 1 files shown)


jgadsden:
do not worry about the failed docker build, this is due to a lack of permissions and not to the change

jgadsden (approved):
thanks for fixing this @ch-lepp , it is very good to have

ch-lepp:
Thx @jgadsden for reviewing and merging this PR so quickly.

Could you pls also have a look into the discussion I opened [here](https://github.com/OWASP/threat-dragon/discussions/993)?
I am pretty confused about the correct redirect uri...

PR #1000, opened 2024-07-05
fix: debug log displays objects properly

### LINKED ISSUE #999: Several debug logs fail to properly display object data
**Describe the bug**:
In the debug log calls often times objects are not properly stringified:
```
const login = (req, res) => {
    logger.debug('API login request: ' + req);
    ...
}
```

Which results in the log output looking like this:
```
threatdragon-1  | debug: controllers/auth.js: API login request: [object Object] {"service":"threat-dragon","timestamp":"2024-07-05 06:03:23"}
```

**Expected behaviour**:
Complex objects need to be properly stringified before logged.

**Environment**:

- Version: 2.2.0

**To Reproduce**:
Activate debug log.

ch-lepp:
**Summary**:
Fixes #999 

Complex objects properly stringified before given to the logger

### DIFF
--- a/td.server/src/controllers/auth.js
+++ b/td.server/src/controllers/auth.js
@@ -9,7 +9,7 @@ import tokenRepo from '../repositories/token.js';
 const logger = loggerHelper.get('controllers/auth.js');
 
 const login = (req, res) => {
-    logger.debug('API login request: ' + req);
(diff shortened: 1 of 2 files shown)


jgadsden (approved):
good to have this update @ch-lepp , I will go ahead and merge to main branch
it may mean that PR #1002 will need a merge from main - will see what happens

PR #1017, opened 2024-07-13
backward compatible for env vars GITHUB_USE_SEARCH and GITHUB_SEARCH_QUERY

jgadsden:
**Summary**:
it would be good to provide backward compatiblity for env vars GITHUB_USE_SEARCH and GITHUB_SEARCH_QUERY
these are probably in use so we should not break that functionality
Only for Github - the bitbucket and gitlab env vars did not work previously so they will not have been used 

make backward compatible for github search env vars

**Other info**:


ch-lepp (approved):
Changes in the code look good.
It's great also, that you added a lot of missing tests for env vars...

The .gitignore approach is a bit too excessive for my taste. But if you like it that way, it is fine by me.

jgadsden:
thanks @ch-lepp , going ahead with the merge

PR #1039, opened 2024-07-25
Update zh.js

yolylight:
add & fix Chinese translation


jgadsden (approved):
thanks for these updates @yolylight , much appreciated

PR #1047, opened 2024-08-06
Change vertices properties from integer to number so they can be imported

### LINKED ISSUE #1046: Exported model does not comply with the JSON scheme
**Describe the bug**:
Exporting the model may lead to `vertices` in the JSON model that contain decimals, when importing the model it expects an integer and will lead to an error message.

**Expected behaviour**:
Either the vertices must be exported as `integer` (no decimals), or the vertices must change of type in the schema so it is able to import vertices as a `number`.

**Environment**:

- Version: 2.2.0
- Platform: Web App and Desktop App
- OS: Only tested on Windows, but looking at the code, it should apply to all: MacOS / Windows / Linux
- Browser: Chrome, Firefox, Safari

**To Reproduce**:
One of my co-workers managed to export a model that had decimals in the vertices values using the same setup as me (Desktop App, Windows machine, version 2.2.0). It is still unknown what action led to these decimals, I have tried many things but have not found how to reproduce it.

The model contains vertices values where the 'x' or 'y' contains a decimal:

```
"vertices": [
    {
        "x": 739.5,
        "y": 640
    }
]
```

The error message after importing the model:
image

mi0x:
**Summary**:
This Pull Request solved the issue: closes #1046 

I have changed the type in the Schema for the `vertices` types to be able to accept decimals, tested it locally and it succesfully imports a model where the x or y vertices are decimals.

### DIFF
--- a/td.vue/src/service/schema/threat-model-schema.V2.js
+++ b/td.vue/src/service/schema/threat-model-schema.V2.js
@@ -393,11 +393,11 @@ export const schema = {
                                                 'properties': {
                                                     'x': {
                                                         'description': 'The horizontal value of the curve point',
-                                                        'type': 'integer'
+                                                        'type': 'number'
                                                     },
                                                     'y': {
                                                         'description': 'The vertical value of the curve point',
-                                                        'type': 'integer'
+                                                        'type': 'number'
                                                     }
                                                 },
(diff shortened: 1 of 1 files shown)


jgadsden (approved):
thanks for correcting this bu @mikehoogesteger and @mi0x 
I will merge it in

PR #1081, opened 2024-09-10
Updates from dependency reports

jgadsden:
**Summary**:
provide updates responding to dependabot and trivy reports, along with some changes to app unit tests

updates from dependency reports

**Other info**:


jgadsden (commented):
this was applied in pull request #1079

PR #1099, opened 2024-10-04
Migrate to version 2.x antv/x6

### LINKED ISSUE #785: Clicking on data flow does not open it in properties
**Describe the bug**:
Clicking on an existing data flow does not set its properties into the properties section. A double click is needed, but it results in an addition of a new point on the data flow.

**Expected behaviour**:
A single click on the data flow makes its properties accessible in the properties section

**Environment**:

- Version: 2.1.1
- Platform: Desktop App and Web App
- OS: MAC OS / Windows / Linux
- Browser: All

**To Reproduce**:
1. Add a data flow
2. Select another component on the diagram
3. Click on the data flow just created

### LINKED ISSUE #954: Implement mousewheel button dragging for graph panning
The current graph panning method requires a combination of keyboard and mouse inputs, and mousewheel button is redundant with left mouse button.

Users should be able to press down the mousewheel button and move the mouse to pan around the graph.

### LINKED ISSUE #1065: Improve Drag and Drop of Components

I have to select the _text_ of a component in the components view in order to drag and drop it into the canvas.
It is not possible to drag and drop by clicking other parts of the component.

**Environment**
Version 2.2
Windows 11
Desktop App
4K Monitor

I want to select any part of the component for drag and drop.

**Additional context**:

2024-08-22_11h50_31

### LINKED ISSUE #1067: Improve selection of edges and creation of bendpoints

Currently, it does not feel intuitive to select a data flow edge or a trust boundary. I have to select the text or the endpoint in order to access the properties. If i click anywhere else on the line, a bend point is immediately created. Often, I don't want to create a bend point but just select the data flow / trust boundary to edit properties.

 I would prefer a UX similar to Miro. When clicking the edge, the properties are shown. When clicking on the circle between two bend points, a new bend point is created.

**Additional context**:

- In Threat Dragon, clicking on line immediately creates bend point.
2024-08-22_13h07_53

- In Miro, clicking on line selects the line first and enables to edit properties. Bend points can then be created afterwards.
2024-08-22_13h08_30

### LINKED ISSUE #1068: I have to click undo five times to actually undo the action
**Describe the bug**:

When creating a data flow by double clicking a process, I have to execute undo five times in order to actually undo the action.

**Expected behaviour**:

Only clicking once should be enough.

**Environment**:

- Version: 2.2
- Platform: Desktop App
- OS: Windows 11

**To Reproduce**:

1. Create Process Element.
2. Double click on it.
3. Click on undo five times.

2024-08-22_13h19_02

### LINKED ISSUE #1082: Can't move ends of line objects inside of Trust Boundary frame
**Describe the bug**:
When line objects (i.e., Data Flow, Trust Boundary lines) are inside of a Trust Boundary frame, manipulation of the line ends causes the manipulated end to be placed at a position other than where the cursor was on release of the left mouse button.

**Expected behaviour**:
Line ends are placed where the mouse cursor is at the time that the left mouse button is released.

**Environment**:

- Version: 2.2.0.0
- Platform: Desktop App
- OS: Windows

**To Reproduce**:
1) On a STRIDE diagram, drag a Data Flow line to the canvas.
2) Drag a Trust Boundary frame to the canvas and resize it so that the Data Flow line is fully enclosed in the Trust Boundary.
3) Click on one end of the Data Flow line and drag it slightly to another position *within* the Trust Boundary.
4) Note that the Trust Boundary highlights yellow (as if something were being anchored to it) once the line end begins to be repositioned.
5) Note that - upon release of the line end - the end will position somewhere else (it looks like the middle of the frame) within the Trust Boundary.
5) Note that selecting only the Trust Boundary and deleting it will delete both objects.

### LINKED ISSUE #1101: Out of scope datastore displayed incorrectly
**Describe the bug**:
Opening the demo model, the datastore 'Web Application Config' is out of scope, but when first opened it is shown as in scope:

Screenshot 2024-10-06 at 16 35 02

only after selecting this component is it properly displayed as out of scope:

Screenshot 2024-10-06 at 16 35 16

**Expected behaviour**:

out of scope elements should always be displayed as such

**Environment**:

- Version: 2.2.0
- Platform: both Web App / Desktop App
- OS: any MacOS / Windows / Linux
- Browser: any chrome, firefox, safari

**To Reproduce**:

open demo model and display the diagram

does not seem to apply to dataflows

jgadsden:
**Summary**:
first pass at migrating to version 2.x antv/x6
unit tests broken
stencil incomplete
various unfinished features

migrate to version 2.x antv/x6

**Other info**:

* the usability of data flows and trust boundaries is _very much_ better, so closes #1067 and closes #785
* double clicking a component to get a data flow is not intuitive, so has been removed to close #1068 
* closes #1082 because the dataflow is still manageable even after it is wrongly associated with a trust boundary
* closes #954 now that we have better scrolling ability
* closes #1101 by fixing schema errors in v2 demo threat model
* closes #1065 by improving the drag and drop of components from the stencil area


jgadsden:
@lreading feel free to progress this, there is a lot of work still to do
the data flows are now very easy to use so the migration will certainly be worth it

leo-reading-springhealth:
Looking at the tests right now, they're giving me a hard time. :weary: 

I only tested the history implementation, and that seems to work.  I'll want to test all the other functions that are implemented via plugin now, as they may or may not work the same.  I found that calling `graph.history` was returning undefined, `graph.getPlugin(<name>)` gets the history plugin which seems to have the same interface as the old version.

leo-reading-springhealth:
Wait, let me try again - I had my wires crossed.  I'm going to start anew from this branch! I had created a new branch and copied some of the work from your old PR.  This is even better, thank you @jgadsden !

leo-reading-springhealth:
Do we want to delete the trust boundary curve and flow stencils?

jgadsden:
Many thanks for the additions @lreading 
I agree we need to have flows and TB curves in the stencils, they are just commented out for now because (I think) they are handled differently in the new stencil package - but I probably have that wrong
There is still a lot of work to do, scrolling the diagram produces very unexpected results :)
The blocking bug where only part of the diagram is displayed has been fixed, but reappears when the diagram is saved, and so on

jgadsden:
> Do we want to delete the trust boundary curve and flow stencils?

definitely want to keep trust boundary curve and flow stencils, I just could not see how to migrate them when I committed these changes

lreading:
There's a bug when clicking undo very quickly after populating the history stack.  No fix for it yet, but I wanted to comment on it so that I don't forget to look into it more. :) 

image

jgadsden:
> There's a bug when clicking undo very quickly after populating the history stack.

yes, agreed - I am fairly sure this is because of an event handler shall I take a look?

td.vue/src/components/GraphButtons.vue
```diff
@@ -87,20 +87,30 @@ export default {
             return;
         },
         undo() {
-            if (this.graph.canUndo()) {
-                this.graph.undo();
+            if (this.graph.getPlugin('history').canUndo()) {
+                this.graph.getPlugin('history').undo();
```
jgadsden, line 91: are we sure that we need the getPlugin() @lreading ? It works with or without using it

leo-reading-springhealth: It is not working on my machine without the `getPlugin`.  The history property is always undefined. :thinking: 
I wonder if there's somehow something different about our environments?

jgadsden: good point, I am on MacOS/ARM64 so it may well be different
it works for me with and without call to getPlugin so I agree we leave it as you have it now

jgadsden:
rubber banding is not working - possibly a conflict between panning and scroller:

```
When using the Scroller plugin at the same time, please do not enable the canvas panning configuration at the same time, because the two forms conflict in interaction.
```

lreading:
I think I have the Flow / Trust Boundaries working as expected now.  They are using their original style.  A Flow can now be added by hovering over any node and dragging one anchor point to another.  When that edge is added, it is converted to a Flow so that we can use the original logic/styles.  I've added threats and played with the properties on the flows, and they seem to work.

Work that's left (please let me know if there is more!)
- Unit tests: fixing and coverage
- Rubber banding(?)
- Manual testing the Electron build (I can do macos and linux, I don't currently have access to a windows PC)

jgadsden:
thanks @lreading , this is a huge improvement and the ability to create data flows from ports on the diagram components is a great improvement, it will be really appreciated
rubberbanding is now OK - it might have quirks but I think it is the best solution; happy for it to be improved

jgadsden:
The electron version behaves as expected, so all good there, although I also do not have access to a windows machine - that is still to be tried
freeze/unfreeze had to be removed as it is only in antv/x6 version 1.x, so I do not know how to do this for the report diagram - will try and look at this over the next couple of days
I can also work on the unit tests

jgadsden:
The only thing left is the unit tests which are in progress and a bug where the the diagram in the report is still modifiable - freeze/unfreeze was removed from antv/x6 1.x -> 2.x and the only reference I can find is something to do with graph.async boolean, and this does not seem to be effective

@lreading I can work on the unit tests and raise an issue for the mutable report diagram - unless you want to determine how it can be done

lreading:
It looks like setting `interacting: false` on the readonly graph gets us really close to an immutable copy for reporting. The cursor will still change to a pointer when hovering over components, but you cannot move them or reconnect edges.  I think this will be good enough? :smile: 

I'm going to take a peek at the unit tests as well!

jgadsden:
wondering if we should provide more ports along the edges of the node components? We have one each side centred , which is a 'must have' , but how about 3 per side to provide less bunching?

ie something like in ports.js : 

```
    items: [
        {
	    group: 'top'
        },
        {
	    group: 'top'
        },
        {
	     group: 'top'
        },
        {
             group: 'right'
        },
 ```
 
to give:

Screenshot 2024-10-14 at 18 51 25

**Comment by lreading** [comment]
> wondering if we should provide more ports along the edges of the node components? We have one each side centred , which is a 'must have' , but how about 3 per side to provide less bunching?
> 
> ie something like in ports.js :
> 
> ```
>     items: [
>         {
> 	    group: 'top'
>         },
>         {
> 	    group: 'top'
>         },
>         {
> 	     group: 'top'
>         },
>         {
>              group: 'right'
>         },
> ```
> 
> to give:
> 
> Screenshot 2024-10-14 at 18 51 25

I think that'd be a great addition!

**Comment by lreading** [comment]
The unit tests are now passing, and it looks like the e2e tests did as well! :tada: 

The coverage is probably down from where it once was, I was primarily focused on getting them passing.  :)

**Comment by jgadsden** [comment]
thanks @lreading for all this - we are ready to commit it I think? The honour is yours :)
maybe I can sneak in one last update to the number of ports ...

File: td.vue/package.json [hunk 2]
```diff
@@ -33,6 +33,7 @@
     "test:e2e-smokes:local": "vue-cli-service test:e2e -C e2e.smokes.local.config.js --url http://localhost:8080/",
     "test:e2e-nightly": "browserstack-cypress run --cf browserstack.nightly.json --sync",
     "test:unit": "vue-cli-service test:unit --testPathIgnorePatterns tests/unit/desktop/",
+    "test:leo": "vue-cli-service test:unit graphMeta.spec.js",
```
lreading: I did not mean to check this in.  That's embarassing. :grin:

td.vue/src/service/x6/ports.js
```diff
@@ -0,0 +1,102 @@
...
+        {
+            group: 'bottom'
+        },
+        {
+            group: 'left'
+        }
+    ]
+};
+
+Graph.registerNode(
```
lreading: I think this was left over from testing, we don't reference customRect anywhere.

lreading (approved):

lreading:
I'll leave it up to you @jgadsden !  If you feel it's ready for merge, by all means :raised_hands:

jgadsden:
apologies @lreading , I was sneaking in a last minute change which removes the ports from the text box - the thinking being that we don't want dataflows attaching to a text box
I also went against the idea of multiple ports on each edge - it did not add to the usability and looked a bit strange
please go ahead and merge, many thanks

PR #1175, opened 2025-01-23
Fix: Add custom About dialog for non-macOS platforms (#421)

### LINKED ISSUE #421: desktop about-box is MacOS only
The desktop drop down menu role: 'about' is probably MacOS only

Use [electron about-window](https://www.npmjs.com/package/about-window) and the [given example](https://github.com/rhysd/electron-about-window/tree/master/example)

**Additional context**

Adding in code to make it cross platform was considered, but use of something like this is not as good as `about-window` :

```
      {
        label: 'about Electron',
        // Set menu roles
        role: 'about', // about, this value is only for Mac OS X systems
        // Click event is invalid when the role attribute of click event can be recognized
        click: () => {
          var aboutWin = new BrowserWindow({ width: 300, height: 200, parent: win, modal: true });
          aboutWin.loadFile('about.html');
        }
      },
```

SidheshwarSarangal:
**Summary**:
This pull request addresses Issue #421, which involves fixing the behavior of the "About" menu item in the application. Currently, the "About" dialog is not functioning as expected on non-macOS platforms. The fix ensures that the "About" dialog appears as a modal window for all platforms except macOS, where the native "About" dialog behavior is retained.

Fixed the "About" menu behavior: non-macOS platforms now open a custom modal window, while macOS retains the native About dialog.

**Other info**:
This PR closes #421 and ensures the "About" menu item behaves appropriately across different platforms.

For macOS, the menu uses the default system behavior.
For non-macOS platforms, the "About" item opens a custom modal window displaying about.html.
This change ensures a more consistent and expected behavior for the "About" dialog across all platforms.

### DIFF
--- a/td.vue/src/desktop/menu.js
+++ b/td.vue/src/desktop/menu.js
@@ -1,9 +1,40 @@
 'use strict';
 
-import { app, dialog, BrowserWindow } from 'electron';
+import { app, dialog, BrowserWindow, ipcMain } from 'electron';
 import path from 'path';
 import logger from './logger.js';
 import { isMacOS } from './utils.js';
+
+
+
+
+function openAboutWindow({ app, BrowserWindow, ipcMain }) {
+    const aboutWindow = new BrowserWindow({
+        width: 400,
+        height: 300,
+        resizable: false,
+        title: 'About',
+        modal: true,
+        parent: BrowserWindow.getFocusedWindow(), // Open as a modal window
+        show: false,
+    });
+
+    aboutWindow.loadURL('data:text/html;charset=utf-8,' + encodeURIComponent(`
+        <h1>About</h1>
+        <p>${app.getName()}</p>
+        <p>App version: ${app.getVersion()}</p>
+        
+    `));
+
+    aboutWindow.once('ready-to-show', () => {
+        aboutWindow.show();
+    });
+
+    ipcMain.on('close-about', () => {
+        aboutWindow.close();
+    });
+}
+
 
 const { shell } = require('electron');
 const fs = require('fs');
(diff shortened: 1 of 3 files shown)


jgadsden (changes requested):
hello @SidheshwarSarangal , thanks for this update, it is good to have
could you run the unit tests with `npm test` and make sure they pass?

td.vue/src/desktop/menu.js
```diff
@@ -153,6 +153,21 @@ export function getMenuTemplate () {
...
+                    label: 'about Electron', // New "About Electron" label
+                    click: () => {
+                        const aboutWin = new BrowserWindow({
+                            width: 300,
+                            height: 200,
+                            modal: true,
+                            parent: mainWindow, // Use mainWindow as the parent if applicable
+                            resizable: false,
+                        });
+                        aboutWin.loadFile('about.html'); // Load the 'about.html' file in the modal
```
jgadsden, line 166: does the `about.html` need to be created?

SidheshwarSarangal: I have checked. Currently, there is no about.html file in the repository. Should I create one for this task? If so, could you confirm the expected content or structure?

SidheshwarSarangal:
I have added the about content. Please tell me if there is anything else.

jgadsden (changes requested):
I think there may be a better way to implement this, for example electron-about-window, or other ways of achieving this cross platform.
This is a commonly needed feature across most electron applications so there is probably going to be an easy path somewhere

SidheshwarSarangal:
Sir, I have used about-window for non mac os

jgadsden (changes requested):
hello @SidheshwarSarangal , could you look at your changes again? they do not seem to use package electron-about-window, see [electron-about-window main.js](https://github.com/rhysd/electron-about-window/blob/master/example/main.js) for an example

jgadsden:
closing this because we need to use use package [electron-about-window](https://www.npmjs.com/package/electron-about-window), referring to [electron-about-window main.js](https://github.com/rhysd/electron-about-window/blob/master/example/main.js) as an example implementation

jgadsden:
closing this because we need to use use package [electron-about-window](https://www.npmjs.com/package/electron-about-window), referring to [electron-about-window main.js](https://github.com/rhysd/electron-about-window/blob/master/example/main.js) as an example implementation

PR #1178, opened 2025-01-27
fix broken repository search

falconfs:
Summary:
This pull request fixes an issue with the repository search functionality when performing threat modeling in an existing repository connected to a GitLab server with many repositories. The issue caused the search results to display a blank page if the desired repository was on a second or subsequent page.

Description for the changelog:
Fix repository search functionality to dynamically fetch results from the Git provider after 500ms when the search query changes.

Other info:
This fix ensures that every time the search input is updated, a new API call is made with the updated search query after a 500ms debounce. The updated results are then fetched from the Git provider, ensuring that repositories are displayed correctly regardless of their pagination.


jgadsden (approved):
thanks @falconfs for this bug fix

jgadsden:
going ahead with the merge - thanks again

PR #1181, opened 2025-01-29
Feat: Add dialog for creating branches and indicators for protected branches

falconfs:
Summary:
This pull request introduces a dialog that allows users to create a new branch during the branch selection process. The newly created branch can then be used directly within the workflow.

Background:
Our repositories have protected branches, which means that users must manually create new branches in the repository before they can work on them. To improve usability, this feature enables branch creation directly from the UI, eliminating the need to navigate to the repository. In addition, the UI now shows which branches are protected and which are not, so users can see in advance if they can make changes to a selected branch.

Description for the changelog:
Added a dialog for creating new branches during branch selection and an indicator for protected branches.

Other info:
All corresponding translations for supported languages have been provided, but they were generated using translation software. Please review for accuracy and completeness.
Displaying protected branches in Bitbucket was unfortunately not straightforward to implement and is therefore not supported in this update.

Screenshot:
image

image


jgadsden (approved):
very good to have this @falconfs , I will go ahead and merge it in

Issue #385, opened 2022-02-14
Resize Diagram or Export to PNG


RockRunner007:
**Describe what problem your feature request solves**
I don't currently have a way to share a large threat model with non-git stakeholders.

**Describe the solution you'd like**
As a Cloud Security Engineer, I would like the ability to export a threat model outside of threat dragon so that I can share a diagram with stakeholders that are not currently using threat dragon.

**Additional context**
MVP - Desktop version could use a resize function 
Long term - Exports options for different formats

lreading:
Hey @RockRunner007 

Thanks for opening the issue!  Version 2 is currently in development and has implemented the zoom feature.  Exporting to PNG/JPEG/SVG is  on the roadmap for v2, but may not be in the initial 2.0 release.  There are improvements in the "report" area, though I can't say for certain if really large models will scale well.  This can be printed or exported to PDF (via print dialogue), which includes the diagrams.

Would the ability to zoom (pan/scroll as well, of course) and/or the report suit your needs, or are there other use-cases you have for the image format?

I'm really interested to know if the scaling is working well with large models in v2.  Would you be willing test one of your larger models using the latest from main?

Install pnpm: `npm i -g pnpm`
Use node 16
From the `td.vue` directory: `pnpm i` then `pnpm run electron:serve`
You can use the "local session" open, "Import a Threat Model via JSON"
You will be presented with an upgrade notification screen, and after that you can navigate to the report directly.  Clicking print then selecting save to PDF from the print dialogue will create the report.

V2 is still under development, so not all features are complete.  The above suggestion will not modify any files on disk. :smile:

RockRunner007:
Hey @lreading 

Thanks for the quick response! 

Deep Dive:
For me personally, the workaround was to use "Toggle Developer Tools" and manually modify the px of the working area so that I could snag a screenshot on my MacBook Pro. The screenshot was imported into a draw.io template that can be shared with leadership and spoken about during a presentation with upper management. For git users, they are accessing the json output via gitlab and modifying mid-sprint. 

Context: https://www.linkedin.com/pulse/threat-modeling-dragon-gitlab-steven-carlson/

Proposed Fix(s):
a. The ability to zoom or expand the working div "tmt-diagram-container" would help me in the short term. I can try to do some testing over the weekend. 
b. V2 of 2.0 would be more ideal as I will be importing the output into templates for sharing outside of my team.

RockRunner007:
[...]

PR #1192, opened 2025-02-07
Add Graph Export to PNG, SVG, JPEG

lreading:
**Summary**:
Adds the ability to export the graph to PNG, SVG and JPEG #385 

Graphs can be exported to PNG, SVG and JPEG.  The SVG export has a transparent background, whereas PNG and JPEG have a white background.

**Other info**:
image

This PR also removes the white background from Flow and Threat Boundary Curve labels.  The only downside I can see is readability in really busy graphs, looking for feedback on this!


jgadsden (approved):
an elegant solution, many thanks @lreading , good to have this for verision 2.4

PR #1197, opened 2025-02-10
clean up console of warnings and errors

### LINKED ISSUE #1100: Bug: Added non-passive event listener to a scroll-blocking event
**Describe the bug**:
there are violations reported in the developer console window when developer tools are enabled, for example:

```
util.js:12 [Violation] Added non-passive event listener to a scroll-blocking 'mousewheel' event.
Consider marking event handler as 'passive' to make the page more responsive.
See https://www.chromestatus.com/feature/5745543795965952
```
and similarly for 'touchstart' events

these are due to the antv/x6 common library, and we need to wait for a newer version of @antv/x6

**Expected behaviour**:

no warnings that slow down the application

**Environment**:

- Version: latest antv/x6 version 2.18.1
- Platform: both Web App / Desktop App
- OS: probably all MacOS / Windows / Linux
- Browser: any chrome, firefox, safari

**To Reproduce**:
open a diagram with developer tools enabled

Screenshot 2024-10-06 at 10 20 14

jgadsden:
**Summary**:
Provides passive events support to touchstart and mousewheel events in the diagram, which prevents the warnings :

- "Added non-passive event listener to a scroll-blocking 'mousewheel' event."
- "Added non-passive event listener to a scroll-blocking 'touchstart' event."

also removes client webSocketURL from Vue devServer config due to error on the console for the desktop app:

- "WebSocket connection to 'wss://localhost/ws' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED"

clean up console of warnings and errors

**Other info**:
closes #1100

### DIFF
--- a/td.vue/vue.config.js
+++ b/td.vue/vue.config.js
@@ -30,13 +30,7 @@ const devServerConfig = hasTlsCredentials
         allowedHosts: [appHostname],
     }
     : {
-        client: {
-            webSocketURL: {
-                protocol: 'wss', // Use secure WebSocket protocol
-                hostname: appHostname,
-                port: 443, // HTTPS port
-            },
-        },
+        // note that client webSocketURL config has been removed, as it was incompatible with desktop version
         port: 8080,
         proxy: {
             '^/api': {
--- a/td.vue/jest.config.js
+++ b/td.vue/jest.config.js
@@ -22,7 +22,7 @@ module.exports = async () => {
         resetMocks: true,
         restoreMocks: true,
         transformIgnorePatterns: [
-            '<rootDir>/node_modules/(?!lodash-es|axios)'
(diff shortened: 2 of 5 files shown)


td.vue/vue.config.js
```diff
@@ -30,13 +30,7 @@ const devServerConfig = hasTlsCredentials
         allowedHosts: [appHostname],
     }
     : {
-        client: {
```
jgadsden, line 33: @ericfitz and @lreading , is this OK to remove? It was the cause of console errors for the desktop app - but seems a bit brutal of me just to delete it

lreading: I believe it's fine to delete, unless @ericfitz disagrees! :)

lreading (approved):

ericfitz:
Sorry I just saw this, this doesn't break anything that I did.

PR #1204, opened 2025-02-14
Cleaning up example.env

ericfitz:
**Summary**:
Making example.env generic

In a previous change, I accidentally leaked a real domain name into example.env.   This change does the following:

- makes the TLS configuration generic
- removes some extra whitespace


jgadsden (approved):
thanks for catching this @ericfitz , I missed it first time round

PR #1205, opened 2025-02-14
Make edge labels legible and selectable

### LINKED ISSUE #1174: Select "data flow" does not automatically open properties window
**Describe the bug**:
When selecting a data flow object, it does not show the properties directly (as for example with a process object) but you have to very cautiously select a certain tiny area on the already selected object and double click to also load the properties in the edit fields below.

**Expected behaviour**:
As soon as the object gets selected, the edit properties form at the lower part of the UI is automatically loaded with its values.

**Environment**:

- Version: all versions
- Platform: local deployment
- OS: tried on Linux  
- Browser: chrome and firefox, same behaviour

**To Reproduce**:
1. Create new model
2. Place "data flow" object
3. Set properties
4. Place other object
5. Try to go back to Object from 2. and modify existing properties, it's a pain

### LINKED ISSUE #1198: Flow labels and trust boundary labels are partially obscured
**Describe the bug**:
Flow labels and trust boundary labels are partially obscured

**Expected behaviour**:
the labels whould have some white space so that they are legible

**Environment**:

- Version: 2.3
- Platform: both Web App and Desktop App
- OS: all MacOS / Windows / Linux
- Browser: all chrome, firefox, safari

**To Reproduce**:

- creata a data flow
- create a trust boundary
- observe labels are partially obscured

Image

jgadsden:
**Summary**:
The labels for  data flow and trust boundary curve can not be used to select the edge - it would be _**very**_ useful if they could
also the labels can be obscured by the line passing through them - this needs whitespace around the label

flow and trust boundary labels made legible and selectable

**Other info**:
also applies a select box around the flow stencil and trust boundary curve stencil, for easier selection of these stencils
closes #1198 
although it does not solve #1174 , this is the best we can do. If the verticex change event is used to select the data flow then the vertex no longer becomes available for change. This comes from AntV/X6 and is unlikely to change
closes #1174

### DIFF
--- a/td.vue/src/service/demo/v2-new-model.js
+++ b/td.vue/src/service/demo/v2-new-model.js
@@ -8,17 +8,7 @@ export default {
     },
     'detail': {
         'contributors': [],
-        'diagrams': [
-            {
-                'id': 0,
-                'title': 'New generic diagram',
-                'diagramType': 'Generic',
-                'placeholder': 'New generic diagram description',
-                'thumbnail': './public/content/images/thumbnail.stride.jpg',
-                'version': '2.3.0',
-                'cells': []
-            }
-        ],
+        'diagrams': [],
         'diagramTop': 0,
         'reviewer': '',
         'threatTop': 0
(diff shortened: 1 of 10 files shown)


lreading (approved):
Lots of great work in here, thank you @jgadsden! <3

PR #1206, opened 2025-02-14
consolidating hostname variable

ericfitz:
**Summary**:
When we added native TLS support to Threat Dragon, we used two environment variables for the host name - one for the app to use and one for TLS certificate validation.  This is unnecessarily complex and error prone, and we have not identified a use case where the names need to be different, so we're consolidating into a single environment variable.

Remove APP_HOSTNAME_TLS, just use APP_HOSTNAME instead


jgadsden (approved):
yes, certainly makes sense to change this - please go ahead and merge @ericfitz

PR #1222, opened 2025-02-27
Fixing label issue for flows and boundary curves

### LINKED ISSUE #1221: Incorrect Labels Shown on Flows and Trust Boundary Curves
**Describe the bug**:
Labels on Flows and Trust Boundary Curves are updated to be the value of the last one touched and not retaining their original value.

**Expected behaviour**:
Flows/Trust Boundary Curves retain their value unless changed

**Environment**:

- Version: latest
- Platform: Web App
- OS: Linux
- Browser: Vivaldi (Chromium)

**To Reproduce**:
- Create a new threat model
-  Add a diagram
- Create a trust boundary curve (or flow) and give it a label of "1"
- Create another trust boundary curve (for flow) and give it a label of "2"
- Save the model
- Close the model
- Re-open the model

Note that BOTH of the boundaries or flows now have a label of "2"

lreading:
**Summary**:
closes #1221 

Bugfix: Data Flows and Trust Boundary Curves showing incorrect labels

**Other info**:
This was identified and raised by the famous @jgadsden :tada: 

I'm not exactly sure why this behavior changed, but I _believe_ it has to do with the antv/x6 upgrade.  Looking at the threat model JSON, I could see that `labels[0].attrs.labelText.text` always had the same text value as the last item that was updated of the same type.  Removing the redundant call to setLabels seems to have fixed this.

### DIFF
--- a/td.vue/src/service/migration/diagram.js
+++ b/td.vue/src/service/migration/diagram.js
@@ -32,6 +32,7 @@ const upgradeAndDraw = (diagram, graph) => {
     updated.thumbnail = diagram.thumbnail;
     updated.id = diagram.id;
     updated.diagramType = diagram.diagramType;
+    graph.getCells().forEach((cell) => dataChanged.updateStyleAttrs(cell));
     store.get().dispatch(tmActions.diagramSaved, updated);
     store.get().dispatch(tmActions.stash);
     store.get().dispatch(tmActions.notModified);
@@ -43,7 +44,6 @@ const drawGraph = (diagram, graph) => {
         console.debug('open diagram version: ' + diagram.version);
         diagram.version = appVersion;
         graph.fromJSON(diagram);
-        graph.getCells().forEach((cell) => dataChanged.updateStyleAttrs(cell));
     } else {
         console.debug('upgrade version 1.x diagram');
         upgradeAndDraw(diagram, graph);
--- a/td.vue/src/service/migration/diagram.js
+++ b/td.vue/src/service/migration/diagram.js
@@ -44,6 +44,7 @@ const drawGraph = (diagram, graph) => {
         console.debug('open diagram version: ' + diagram.version);
         diagram.version = appVersion;
         graph.fromJSON(diagram);
+        graph.getCells().forEach((cell) => dataChanged.updateStyleAttrs(cell));
     } else {
         console.debug('upgrade version 1.x diagram');
         upgradeAndDraw(diagram, graph);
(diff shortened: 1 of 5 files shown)


jgadsden:
this is a difficult one for sure - the good news is that it was introduced when I made changes for the selectable / grabbable edge labels with whitespace
the thinking is that it will help select the data flows and trust boundary curves

I did a bit more digging around and the root cause is the global constant `defaultLabel` is being applied across all flows (or boundaries) that AntV knows about ie the ones that have been modified during the session. So when the name in `defaultLabel` is changed then it is applied across all (my bad for sure)

so I _think_ I have it working with for example:

```
TrustBoundaryCurve.prototype.setName = function (name) {
    let newLabel = 	this.getLabels();
    this.setLabels([name]); // updates the label as it is being written
    newLabel[0].attrs.labelText.text = name;
	// set the label so that it can be grabbed and also has white space
    this.setLabels(newLabel);
```

and after deleting the global the define becomes
 
```
// trust boundary curve (edge, dotted line)
export const TrustBoundaryCurve = Shape.Edge.define({
    constructorName: name,
    width: 200,
    height: 100,
    zIndex: 10,
    attrs: {
        line: {
            strokeWidth: 3,
            strokeDasharray: '10 5',
            sourceMarker: null,
            targetMarker: null
        },
        rect: {
            fill: 'none',
            stroke: 'none'
        },
    },
    labels: 	[
	    {
	        markup: [
	            {
	                tagName: 'ellipse',
	                selector: 'labelBody',
	            },
	            {
	                tagName: 'text',
	                selector: 'labelText',
	            },
	        ],
	        attrs: {
	            labelText: {
	                text: '',
	                textAnchor: 'middle',
	                textVerticalAnchor: 'middle',
	            },
	            labelBody: {
	                ref: 'labelText',
	                refRx: '50%',
	                refRy: '60%',
	                fill: '#fff',
	                strokeWidth: 0,
	            },
	        },
	        position: {
	            distance: 0.5,
	            args: {
	                keepGradient: true,
	                ensureLegibility: true,
	            }
	        },
	    }
	],
    connector: 'smooth',
    data: defaultProperties.boundary
});

```

lreading:
It seems you have a far better grasp on this than I do!  Feel free to update the PR or open a new on if you have a better fix! <3

jgadsden:
yes, apologies @lreading , I was being a bit opaque :)

the way I had declared labels for both data flows and trust boundaries meant that any changes to labels were applied globally across all edges that had been modified

this was a bug introduced recently to make the edges easier to grab - clicking on the edge label will also select the edge, before the only way to select the edge was to click on the target arrow

I have committed some suggestions, but please revise / throw away / accept - over to you

lreading:
I understand now! Great catch.  My initial attempt was fixing a symptom of the problem, you found and addressed the root cause.  Great work as always, @jgadsden !!!! :tada:

PR #1227, opened 2025-03-02
Cleanup and adding clarity around port configuration

ericfitz:
**Summary**:
Cleans up vue.config.js to make clear what different "port" values are used for
Adds error checking for reading TLS certificate/key files
Fixes up references to renamed "PORT" variable in server app & tests
Reorganizes and heavily comments example.env 

Make configuration of custom ports easier and more understandable


ericfitz:
I wasn't sure if exit(1) is the right way of handling an exception reading the TLS certificate files - let me know if this needs to be done in a different way.

I also reordered and extensively commented example.env to make it much more understandable.

ericfitz:
Closing due to failing unit test.  Will submit again after fix.

PR #1228, opened 2025-03-03
Clean up, add error checking, remove ambiguity around term "port"

ericfitz:
**Summary**:
Cleans up vue.config.js to make clear what different "port" values are used for
Adds error checking for reading TLS certificate/key files
Fixes up references to renamed "PORT" variable in server app & tests
Reorganizes and heavily comments example.env 

Make configuration of custom ports easier and more understandable

### DIFF
--- a/example.env
+++ b/example.env
@@ -1,36 +1,73 @@
-BITBUCKET_CLIENT_ID=01234567890123456789
-BITBUCKET_CLIENT_SECRET=0123456789abcdef0123456789abcdef0123456
-BITBUCKET_SCOPE=repository:write
-BITBUCKET_WORKSPACE=workspace_name
-
+# these keys are used to protect communication between the application and the server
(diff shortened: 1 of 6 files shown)


ericfitz:
Jon, it appears that I'm dereferencing the "port" variable before use, but I can't see where I'm making a mistake.  Would you take a minute or two to see where?  Then punt this back to me and I will fix the issue.

In the meantime I will keep looking and I'll close and resubmit when I fix the issue.

jgadsden (approved):
very good to have these changes @ericfitz , and very good to see the annotation to the example.env file
I am working on the documentation that may help as well
we have just released version 2.4.0, with version 2.5.0 scheduled 6 months away, but this is well worth putting into a version 2.4.1 which can potentially be released this month (March 2025) or mid next month

lreading (approved):
Awesome stuff, thanks @ericfitz :tada:

PR #1255, opened 2025-04-09
Restore documentation to threat dragon container

jgadsden:
**Summary**:
This brings the version 2.x Threat Dragon documentation back into the source repo

restore documentation to threat dragon container

**Declaration**:

- [x] any use of AI is declared in this pull request

**Other info**:


jgadsden:
Hello @lreading , here is the docs reinstated to the pages at https://www.threatdragon.com/docs/
it also uses github.com/lreading/owasp-td-jekyll , if you wanted to unarchive this repo?

docs/Gemfile
```diff
@@ -7,17 +7,17 @@ source "https://rubygems.org"
 #
 # This will help ensure the proper Jekyll version is running.
 # Happy Jekylling!
-gem "jekyll", "~> 4.3.2"
+gem "jekyll", "~> 4.3.4"
 
-# This is the default theme for new Jekyll sites. You may change this to anything you like.
-gem "minima", "~> 2.5"
+# Theme
+gem "owasp-td-jekyll", "~> 0.2.1"
```
lreading, line 13: This was last updated 4 years ago (man, my grays are starting to show).  I had archived it where we weren't using it.  Would it make sense for me to do a gemfile update and publish a new version?

https://github.com/lreading/owasp-td-jekyll

lreading: Looks like you and I were on the same page @jgadsden !

I've unarchived the repo and will see what I can do about updating the gemfile.  I can only assume it needs an update.  Given that the ouput is a static website, I think it's probably fine to continue without the updates?  I'll leave that up to you!

lreading: I pushed an update: https://rubygems.org/gems/owasp-td-jekyll 
The newest version is 0.2.2 if we'd like to use that one.  The only difference is updated dependencies.  I ran the theme locally and all looked good to me. :sunglasses:

lreading (approved):
This was a lot of work, thank you so much for it!  I'm excited to have the docs on an official OWASP domain as well!

I'm leaving it up to you if you want to bump the version of the theme or not as part of this or if we want to do that later.  This is a big enough PR :sweat_smile:

docs/Gemfile
lreading: The gemfile.lock was deleted, so I updated the .gitignore to include it and pushed it with the change to the Gemfile

jgadsden:
Hej hej @lreading 
I could not find out how to stop Trivy from complaining about example secrets - if you can find a way, I bow to you

.github/workflows/pull_request.yaml
```diff
@@ -354,3 +354,4 @@ jobs:
           format: 'table'
           trivyignores: '.github/workflows/.trivyignore'
           exit-code: 1
+          skip-files: '/app/docs/configure/bitbucket.html'
```
lreading, line 357: @jgadsden - I did _technically_ find a way to get it to skip the file in question.  

I'm not a huge fan of skipping the file entirely, or the potential of the list of skipped files growing in the future.  That said, it is a static file for our documentation, so I feel that the risk is low enough and we can keep getting the green light from our Trivy scan.  Let me know if you're on board with this or if you'd like me to continue exploring alternatives. :)

jgadsden: I agree with your solution - we should skip the docs files
I have had feedback in the past that having the env vars spelt out has been very useful, and that is what Trivy is complaining about (or at least the ones that say secrets)
let's go with this solution and merge

after a two year gap, we are back to nicely formatted docs 🥇

PR #1321, opened 2025-07-06
Document and revise TM-BOM import

jgadsden:
**Summary**:
This started out as documenting the TM-BOM schema used by Threat Dragon
and then thi identified gaps in the TM-BOM import which have been corrected

document and revise TM-BOM import

**Declaration**:

- [x] any use of AI has been declared in this pull request

**Other info**:


lreading (approved):
I think that makes a lot of sense to have the toast message. :) It's similar to what we did when we enabled the auto upgrade from v1-v2, and makes it clear that it's still a bit of a work in progress.

Issue #1360, opened 2025-10-17
Bug: loss of diagram changes for a reopened model


kfabjans:
**Describe the bug**:
When using the Menu Bar options Save or Save As in OWASP Threat Dragon, the threat modeling diagram is not saved. This can lead to significant data loss, especially for teams collaborating on models.
However, using the built-in Save button (next to the Close button in the app) correctly saves the current work.

**Expected behaviour**:
Selecting Save or Save As from the menu bar should persist the current threat model diagram to the chosen file or location, just like the built-in Save button does.

**Environment**:

- Version: 2.5.0
- Platform: Desktop App
- OS: macOS 26.0.1 (Tahoe) and Windows 11
- Browser: N/A (Desktop app issue)

**To Reproduce**:

1. Open OWASP Threat Dragon Desktop App.
2. Create or edit a threat modeling diagram.
3. From the Menu Bar, click File → Save or File → Save As.
4. Close and reopen the diagram.
5. Observe that changes were not saved.

**Any additional context, screenshots, etc**:

<img width="3828" height="278" alt="Image" src="https://github.com/user-attachments/assets/9c49d94e-64b1-44e8-8f3a-a8579045b31f" />

jgadsden:
Hello @kfabjans , can you confirm that you are using the latest version: v2.5.0
This works as expected because it was the subject of PR #1331 

Can you provide a step-by-step of how you make it not do as expected?

PR #1376, opened 2025-11-04
Docs: Investigation and proposed fix for menu Save and Save As issue (#1360)

Awaneesh03:
### Summary
This PR documents the investigation and proposed fix for issue [#1360](https://github.com/OWASP/threat-dragon/issues/1360).

### Context
In the Threat Dragon desktop app, the File → Save and File → Save As menu options do not persist diagram changes.  
The in-app Save button works correctly because it triggers an IPC event in the renderer process that writes the updated model to disk.  
The menu handlers currently do not send that event, so the changes are lost when reopening the diagram.

### Proposed Fix
Update the Electron main process menu definitions to send the same events used by the in-app Save button:

```js
{
  label: 'Save',
  accelerator: 'CmdOrCtrl+S',
  click: () => mainWindow.webContents.send('save-model')
},
{
  label: 'Save As',
  accelerator: 'CmdOrCtrl+Shift+S',
  click: () => mainWindow.webContents.send('save-model-as')
}

### DIFF
--- a/docs/fix-save-menu-issue.md
+++ b/docs/fix-save-menu-issue.md
@@ -1,0 +1,39 @@
+# Investigation: Menu bar "Save" and "Save As" do not persist the threat model diagram
+
+**Related issue:** [OWASP/threat-dragon#1360](https://github.com/OWASP/threat-dragon/issues/1360)
+
+## Summary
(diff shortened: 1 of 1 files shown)


jgadsden (changes requested):
Thank you for the pull request @Awaneesh03 , but pull request discussions are not part of the documentation set

Issue #1388, opened 2025-11-19
Add Trackpad Pinch-to-Zoom Support (Like Canva/Figma)


ayushjava07:
---

Describe what problem your feature request solves:

Currently, Threat Dragon does not support touchpad pinch-to-zoom gestures (two-finger zooming).
When diagrams become large, it's very difficult to zoom in/out smoothly using only:
Ctrl + mouse wheel
Zoom buttons
For laptop users (Linux/Mac/Windows), navigation becomes slow and uncomfortable.
Modern tools like Canva, Figma, Miro all support smooth pinch zoom, which improves usability a lot.

---

Describe the solution you'd like:
Add native pinch-to-zoom support for touchpad users.
This means:
Two-finger pinch-out → Zoom in
Two-finger pinch-in → Zoom out
Smooth zoom transitions
Works in both Web and Desktop versions
This should behave similar to Canva/Figma-style viewport zooming for smooth diagram navigation.

---
Additional context:

I tested this on Threat Dragon v2.5.0 (desktop + web)

Trackpad gesture does not zoom the canvas

I have attached the screenshot of the diagram area
Describe what problem your feature request solves:

Currently, Threat Dragon does not support touchpad pinch-to-zoom gestures (two-finger zooming).
When diagrams become large, it's very difficult to zoom in/out smoothly using only:

Ctrl + mouse wheel
Zoom buttons

For laptop users (Linux/Mac/Windows), navigation becomes slow and uncomfortable.
Modern tools like Canva, Figma, Miro all support smooth pinch zoom, which improves usability a lot.

 ---
Describe the solution you'd like:

Add native pinch-to-zoom support for touchpad users.

This means:

Two-finger pinch-out → Zoom in
Two-finger pinch-in → Zoom out
Smooth zoom transitions
Works in both Web and Desktop versions
This should behave similar to Canva/Figma-style viewport zooming for smooth diagram navigation.

Additional context:

I tested this on Threat Dragon v2.5.0 (desktop + web)
Trackpad gesture does not zoom the canvas
If approved, I am willing to test or help implement this feature

---

jgadsden:
Hello @ayushjava07 , what is the use case for using Threat Dragon in this way?
It is very unlikely to be used on a mobile phone or a tablet

ayushjava07:
hello @jgadsden , got it .

PR #1391, opened 2025-11-23
Add consistent pinch-to-zoom support for the diagram editor

ankitBytes:
# PR: Add consistent pinch-to-zoom support for the diagram editor

## Overview
This update adds proper pinch-to-zoom support to the Threat Dragon diagram editor.  
Before this change, zooming only worked on machines where trackpad pinch gestures sent `ctrlKey = true` (mostly macOS).  
On many Windows and Linux laptops, pinch gestures arrived as normal wheel events, so the graph never zoomed.

This PR makes pinch-to-zoom behave the same across all devices.

---

## What was happening
Different systems send different wheel event patterns when the user performs a pinch gesture:

- **macOS:** `wheel + ctrlKey = true`
- **Windows / Linux:** large `deltaY`, but `ctrlKey = false`
- **Touch devices / emulation:** varying wheel deltas
- **VMs:** simple wheel events only

Because X6’s built-in zoom depended on the `ctrl` modifier, most machines couldn’t pinch-zoom at all.

---

## What I changed
All changes are inside `getEditGraph()` in `graph.js`:

- Disabled X6’s default mousewheel zoom to avoid conflicts  
- Added a small custom pinch detector that triggers when:
  - `ctrlKey` is present **or**
  - `deltaY` is large enough to match Windows/Linux pinch behaviour
- Applied zoom using `graph.scale()` for stable, absolute zooming  
- Added zoom bounds (`0.5` to `3`)
- Adjusted the zoom sensitivity to make pinch feel smoother  

No other files or UI components were touched.

---

## Testing done
I tested the behavior across several environments:

- Chrome device simulation (iPad, Android tablet, touch laptop, desktop)
- A laptop trackpad with native pinch support
- Virtual machines (Windows and Linux)
- Mouse wheel scrolling

Pinch-to-zoom works consistently in all cases, and regular scrolling does not trigger zoom.

---

## Why this change is safe
- The update only adds custom zoom handling and disables X6’s built-in wheel zoom  
- Diagram data, save logic, selection, and node/edge interactions remain unchanged  
- All modifications are limited to `graph.js`  

---

## Issue linked
Fixes: **#1388 – Add Trackpad Pinch-to-Zoom Support**

https://github.com/user-attachments/assets/4894e112-8582-4ca1-bbf5-fd564c02625f

Note: AI was used to help draft parts of this PR description, but all code and testing were done manually.

### DIFF
--- a/td.vue/tests/unit/service/x6/graph/graph.spec.js
+++ b/td.vue/tests/unit/service/x6/graph/graph.spec.js
@@ -14,15 +14,25 @@ describe('service/x6/graph/graph.js', () => {
         use = jest.fn().mockReturnThis();
     }
 
+    function createMockContainer() {
+        return {
+            foo: 'bar',
+            addEventListener: jest.fn(),
+            removeEventListener: jest.fn(),
+            dispatchEvent: jest.fn(),
+        };
+    }
+
     describe('getReadonlyGraph', () => {
         beforeEach(() => {
-            container = { foo: 'bar' };
+            container = createMockContainer();
             events.listen = jest.fn();
             graph.getReadonlyGraph(container, GraphMock);
         });
 
         it('does not add the event listeners', () => {
             expect(events.listen).not.toHaveBeenCalled();
+            expect(container.addEventListener).not.toHaveBeenCalled();
         });
     });
 
@@ -30,7 +40,7 @@ describe('service/x6/graph/graph.js', () => {
         let graphRes;
 
         beforeEach(() => {
-            container = { foo: 'bar' };
+            container = createMockContainer();
             events.listen = jest.fn();
             keys.bind = jest.fn();
             graphRes = graph.getEditGraph(container, GraphMock);
@@ -126,9 +136,9 @@ describe('service/x6/graph/graph.js', () => {
 
         it('enables the mouse wheel', () => {
             expect(graphRes.mousewheel).toEqual({
-                enabled: true,
-                global: true,
-                modifiers: ['ctrl', 'meta']
+                enabled: false,
+                global: undefined,
+                modifiers: undefined
             });
         });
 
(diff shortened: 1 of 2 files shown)


ankitBytes:
Thanks for the heads up!
I didn't intentionally updated any dependencies the `package-lock.json`  changes came from running `npm install` during setup. Now, I have also revert it back to the original lockfile, so that the PR stays cleaner.

td.vue/package.json
```diff
@@ -69,7 +69,7 @@
     "@fortawesome/free-brands-svg-icons": "^6.2.0",
     "@fortawesome/free-solid-svg-icons": "^6.2.0",
     "@fortawesome/vue-fontawesome": "^2.0.8",
-    "ajv": "^8.11.0",
+    "ajv": "^8.17.1",
```
jgadsden: @ankitBytes this change is not needed, the `^` already ensures that the latest version 8.x.x of `ajv` is used
and both `package-lock.json` files should be restored using `git restore`

ankitBytes:
Thanks for the update! I am working on the failing tests and cleaning up the PR. 
I will push the updates soon and mark the PR ready for review again.

td.vue/tests/unit/service/x6/graph/graph.spec.js
```diff
@@ -158,5 +168,14 @@ describe('service/x6/graph/graph.js', () => {
         it('uses the export plugin', () => {
             expect(graphRes.use).toHaveBeenCalledWith(new Export());
         });
+
+        // NEW TEST: wheel event listener added by your feature
```
lreading: Nit-picky: remove this comment, it doesn't add anything, just an AI artifact. :)

ankitBytes:
Hi, just checking in. Is there any update or anything I need to update, then please let me know.

jgadsden (changes requested):
I do not see what this pull request is actually fixing, and I am concerned that we are disabling existing mousewheel zoom functionality, if I understand the changes correctly

We may be changing the user interface functionality that has been in place for years now, and without warning the Threat Dragon community
The existing mousewheel functionality works correctly on all my test systems, so ideally we want to preserve that

It is worth noting that the Threat Dragon community is global and very extensive, ranging from large enterprises of over 100,000, through to small enterprises, and many academic institutions

If we really want to go ahead with this change then we would have to modify the keyboard user instructions and be very clear about what mousewheel functionality has changed

Issue #1373, opened 2025-11-02
Integration between Cornucopia and Threat Dragon


jgadsden:
**Describe what problem your feature request solves**:
Provide Cornucopia and EoP focused threat modeling within OWASP Threat Dragon for easier and more comfortable use of threat elicitation, while playing the game, and a better gaming experience.

**Describe the solution you'd like**:
An API hosted at cornucopia.owasp.org will provide the card codes to build the template for the EoP diagram based on the existing STRIDE diagram type.

**Additional context**:
Possible solution : 
Card codes will be fetched during the OWASP Threat Dragon build process through an API hosted at cornucopia.owasp.org to build the template for the EoP diagram based on the existing STRIDE diagram type. The new diagram will be identical except for that the threat model will contain support for selecting a specific EoP game and the card played.

<img width="1155" height="630" alt="Image" src="https://github.com/user-attachments/assets/6dcd846e-84bc-420c-8fec-fe00d2c526f5" />

jgadsden:
Thanks for the .json @sydseter 
There are many possible implementations, one possibility is that a new diagram type could be defined (for example 'Cornucopia') and instead of the STRIDE categories it could have the [Cornucopia categories](https://cornucopia.owasp.org/cards) drop down list in the Threat dialog:

* Data validation & encoding
* Authentication
* Session management
* Authorization
* Cryptography
* Cornucopia
* Wild card

Then depending on the category chosen a further drop down could select cards within that category, ideally including a thumbnail of the any card selected.
I agree that I may be asking for the moon :)

PR #1403, opened 2025-12-13
Integration between Cornucopia and Threat Dragon

Clarensedor:
**Summary**:

The API hosted at `cornucopia.owasp.org` provides the card codes to build the template for the EoP diagram based on the existing STRIDE diagram type.
The EoP cards can be applied to all types of component, and is not 'per element'

closes: #1373 

integration between Cornucopia and Threat Dragon

**Declaration**:

- [ ] appropriate unit tests have been created / modified
- [ ] functional tests created / modified for changes in functionality
- [ ] any use of AI has been declared in this pull request

**Other info**:
Agregue los package.json para cuando se hace  npm run dev:vue tambien agregue en el gitignore para que no suba estos json


.gitignore
```diff
@@ -290,3 +292,7 @@
 # utility scripts
 !utils
 !utils/**/*
+
+
```
jgadsden, line 296: this .gitignore file is used as an allow-list, rather than deny-list, by denying everything unless it is specifically removed from the list
and as `minimal.env` is already included  on line 34, then this should be:

```
!td.vue/src/service/schema/cornucopia.json
```

It takes some getting used to but it is much more secure - it helps to avoid secrets being included in pull requests (it can happen!)

jgadsden: This is why `minimal.env` is no longer in the fileset - we need it because it is referenced in the documentation
and other files such as `schema/cornucopia.json` are missing, along with `td.vue/src/plugins/generate-cornucopia.ts`

td.vue/package.json
```diff
@@ -5,6 +5,8 @@
   "appBundleId": "org.owasp.threatdragon",
   "private": true,
   "scripts": {
+    "predev:vue": "node td.vue/src/plugins/generate-cornucopia.js",
+    "prebuild:desktop": "ts-node ./src/plugins/generate-cornucopia.ts",
```
jgadsden: TS is used here, and JS is used by "prebuild:desktop" in `../package.json`

td.vue/src/views/ThreatModelEdit.vue
```diff
@@ -320,6 +327,7 @@ export default {
                 || this.model.detail.diagrams[idx].title === this.$t('threatmodel.diagram.linddun.defaultTitle')
                 || this.model.detail.diagrams[idx].title === this.$t('threatmodel.diagram.plot4ai.defaultTitle')
                 || this.model.detail.diagrams[idx].title === this.$t('threatmodel.diagram.stride.defaultTitle')
+                || this.model.detail.diagrams[idx].title === this.$t('threatmodel.diagram.eop.defaultTitle')
```
jgadsden, line 330: could this have an entry in the `i18n` translation files?

Clarensedor:
thanks for the  feedback,  i accindentaly made a PR witch was intended for the fork

PR #1404, opened 2025-12-13
Add AI Tools extension to Threat Dragon

### LINKED ISSUE #1371: AI-Powered Threat Modelling Extension
It would be good to build in a generative AI engine into Threat Dragon that suggests threats for a given data flow diagram model.

The benefits have been described in [AI-Powered Threat Modeling with OWASP Threat Dragon](https://infosecotb.com/ai-powered-threat-modeling-with-owasp-threat-dragon-part-2-generating-threats-with-artificial-intelligence) – Part 2 (along with [Part 1](https://infosecotb.com/using-owasp-threat-dragon-and-artificial-intelligence-ai-for-threat-modelling-part-1-creating-a-data-flow-diagram/))

This [TD AI Modeler tool](https://github.com/InfosecOTB/td-ai-modeler) is open source and @InfosecOTB  has provided [example threats generated](https://github.com/InfosecOTB/td-ai-threat-models) for the threat models

It would be good to integrate this into Threat Dragon, possibly following [ideas on the implementation](https://github.com/OWASP/threat-dragon/discussions/1358#discussioncomment-14786712) from the discussion topic

initial integration with the desktop version

**Additional context**:

This is the over-arching issue with various tasks:

jgadsden:
**Summary**:
This is an implementation of proposing threats for a given threat model using generative AI for the desktop version

This is a long term initiative that s led by @InfosecOTB from discussion : [AI-Powered Threat Modelling Extension](https://github.com/OWASP/threat-dragon/discussions/1358)

AI tools extension for desktop

**Declaration**:

- [x] any use of AI has been declared in this pull request

**Other info**:
This is a draft pull request to allow comment and further refinement
It will close: #1371


InfosecOTB:
E2E tests failed before due to keytar Linux dependencies. Although it passed previously, I’ve replaced it with Electron’s safeStorage for better compatibility.

td.vue/ai-tools/src/ai_client.py
```diff
@@ -0,0 +1,99 @@
...
+    except Exception as e:
+        logger.error(f"Problem with getting max tokens: Using default value of 100000.")
+        max_tokens = 100000
+
+    # Build API completion parameters
+    completion_params = {
+        "model": model_name,
+        "messages": messages,
+        "temperature": temperature,
+        "timeout": 14400,
```
lreading: 4 hours is a pretty long time.  No need to change it right now, but maybe in the future we can make this configurable as well!

td.vue/ai-tools/src/ai_client.py
```diff
@@ -0,0 +1,99 @@
...
+    logger.debug(f"\n\nResponse: {response}")
+    
+    # Parse and validate AI response
+    try:
+        ai_response = AIThreatsResponseList.model_validate_json(response.choices[0].message.content)
+    except Exception:
+        # Fallback: try to extract JSON from markdown or plain text
+        logger.warning("LLM returned invalid JSON. Trying to extract JSON...")
+        # Look for JSON object containing "items" key (our expected format)
+        match = re.search(r'\{\s*"items"\s*:\s*\[.*?\]\s*\}', response.choices[0].message.content, re.S)
```
lreading, line 84: I like this approach!
Sometimes, especially with smaller models, LLMs will start their response with "Ok, ```json" or similar despite being given specific instructions NOT to include markdown formatting in the response.  I think this is a great fallback to avoid those cases!

td.vue/ai-tools/src/main.py
```diff
@@ -0,0 +1,312 @@
...
+    if not settings_path.exists():
+        raise FileNotFoundError(f"Settings file not found: {settings_path}")
+    
+    # Load settings from JSON file
+    with open(settings_path, 'r', encoding='utf-8') as f:
+        settings = json.load(f)
+    
+    # Extract settings with defaults
+    result = {
+        'llm_model': settings.get('llmModel', ''),
```
lreading, line 99: This could be a lack of knowledge on my part as I've never worked with litellm and have only partially read the docs: Will this still work with some LLMs if you do _not_ include an llm_model parameter?  If not, would it make sense to guard this similar to how we did other settings below?

lreading:
This doesn't need to be part of this PR, but one concern I have is around how future model changes may impact the system instructions.  We should consider updating the contributing guidelines to include that schema changes (which should be rare between major version bumps anyways) might require reviewing the system instructions. 

This is an impressive amount of work, and I can't thank you enough!  I'm going to be testing this locally on different devices over the weekend! :)

lreading:
There's nothing actionable for this PR in this comment - I'm posting this for future readers (and maybe future us) who want to use this feature with locally hosted models.

# Notes on Running This With Locally Hosted Models
I want to share some of my notes with future readers who are interested in using this with locally hosted LLMs.  The TLDR is that ***I don't recommend running this with local LLMs, at least right now.*** There were some tweaks I had to make to get the AI Tools integration to work with locally hosted models.  I wanted to share my notes in case anyone decides to go down this route now or in the future.  I do believe that with some refinement to the prompt and future improvements in open source LLM models, running with a local model will one day be perfectly viable.  

@InfosecOTB  describes in their blog post that smaller LLM models didn't work well.  They also mentioned that they were working with limited hardware resources (an office pc, as well as a gaming pc) so they couldn't run huge models.  I have access to a system that can run 30-32b parameter models.  Even with larger models, there were still too many problems for me to recommend this strategy.

# Notes on Running With a Local Model

- All of my testing used the example "Generic CMS" threat model
- I had to update the code to allow the integration to work without an api_token. litellm (or a dependency thereof)  got angry about invalid headers when trying to construct the authorization/bearer token header.
- Everything else worked as expected: set the api base and model name (ollama/whatever-model)
- Locally hosted reasoning models are not returning threats
  - It seems to always return no threats (empty `items: []`), and the inference time doesn't align with what I get when testing manually outside of Threat Dragon.  Also, testing outside of TD, the LLM consistently returns threats
  - The system prompt / request was being truncated to 4096
  - There's potentially a fix by adding ollama-specific completion_params["options"], but I didn't go that deep
- "-instruct" LLM model variants, on average, worked better than the "regular" models
- Some smaller LLM models got stuck in a loop and continued spitting out the same id/threat until it crashed out
- Even with 30-32b param LLM models, hallucination, especially with element ids, was a real problem
  - The model kept creating an id that didn't exist, despite the system prompt telling it not to
  - Weirdly, that hallucinated id was consistent across runs (still no idea why? if anyone has thoughts, please share!)
- Local models did not do well understanding trust boundaries - it would regularly generate threats about flows crossing boundaries, when they did not actually cross any boundaries
- System Prompt Tweaks:
  -  Updating the AIThreatsResponse id parameter to "element_id" in both the python code as well as the system prompt slightly reduced the id hallucination
  - Adding an instruction to consider id/threat title as a compound / unique key resolved the infinite looping problem with the smaller models
  - Passing a list of "available ids" in the prompt seems to resolve the id hallucination issue (code below)
  - I wasn't able to stop the LLM from adding ids into the threat descriptions / mitigations.  It really likes referencing things by id. 

## Passing Available Ids

*The following was AI generated - proceed with caution.*

```python
def extract_available_element_ids(model: Dict[str, Any]) -> Tuple[List[str], Dict[str, Dict[str, Any]]]:
    """
    Returns:
      - sorted list of allowed element IDs (cell.id) for threat attachment
      - element_map: id -> cell (handy for validation / names / shapes)
    """
    element_map: Dict[str, Dict[str, Any]] = {}
    allowed: Set[str] = set()

    diagrams = (((model or {}).get("detail") or {}).get("diagrams") or [])
    for d in diagrams:
        cells = (d or {}).get("cells") or []
        for cell in cells:
            if not isinstance(cell, dict):
                continue

            cid = cell.get("id")
            shape = cell.get("shape")
            if not isinstance(cid, str) or not cid:
                continue
            if shape not in IN_SCOPE_SHAPES:
                continue

            data = cell.get("data") or {}
            out_of_scope = bool((data or {}).get("outOfScope", False))
            if out_of_scope:
                continue

            allowed.add(cid)
            element_map[cid] = cell

    return sorted(allowed), element_map
```

# Tested Models

I think that 30-32b parameter models have potential with the current setup, but there's some work we'd have to do to make it more reliable and actionable.  7b parameter models performed poorly, and were consistently unable to provide meaningful threats.  This could be due to the context window - we are sending a pretty sizable prompt to the LLM by including both the schema and the threat model.  I was only testing this with a _very small_ threat model with a single diagram, It's safe to assume that with larger threat models, 7b param LLM models will fail spectacularly.

|Model|Notes|
|---|---|
|ollama/qwen3:32b|Always returned an empty array.  If I copy the system prompt and feed it to qwen3 outside of Threat Dragon, it would generate threats.  This is *potentially* a configuration/integration issue with litellm and ollama or reasoning models, relating to the truncation issue referenced above.|
|ollama/qwen3:30b-instruct|This was the best model I tested.  **It still requires tweaks to the system prompt to prevent hallucination**.  The output was still less than ideal, but better than any of the other models I tried.  The threats were relevant and associated correctly (after tweaking the prompt) with the exception of understanding where trust boundaries lie.  It would frequently suggest threats that describe crossing trust boundaries, when the flows it referenced do not actually cross any trust boundary in the diagram.  I also couldn't get the LLM to stop referencing ids in the threat descriptions/mitigations.|
|qwen3-coder:30b|This was a close second to the qwen3:30b-instruct model.  It suffered from the same problems with understanding trust boundaries, but tended to lean into trust boundaries even more when generating threats.  In many runs, all of the threats it generated were focused on data crossing trust boundaries.  That would be fine, but it is not actually understanding the trust boundaries correctly, so the output was typically even less useful.|
|qwen2.5-coder:7b|Definitely not up for the task.  Out of 15 attempts, only one passed validation - even at that, it didn't provide any meaningful threats and was full of warnings.|

Of course, there are PLENTY of other models that could be tested, but I think my time will be better spent exploring and manually testing the intended path for this new tooling. :smile: 


I have reviewed the code and left a couple of comments, but there's absolutely nothing that I'd consider a blocker!  My next steps include manual, cross-platform testing using cloud models (as is the intended use-case).  I'm going to aggregate my feedback / thoughts / suggestions based on the manual testing, and present it in a single comment to try and keep the PR somewhat organized.

td.vue/ai-tools/src/ai_client.py
InfosecOTB: Yes - 4 hours is a pretty long time, but it was there for a reason. When I tested local models with Ollama, it sometimes took a few hours to get any result. However, I like your idea of adding this parameter to the settings, so I adjusted the code and set the default timeout to 15 minutes.

td.vue/ai-tools/src/ai_client.py
InfosecOTB: In general, it’s quite difficult to get structured JSON output from some LLMs, so I’ve tried to make it as universal as possible. I’ll provide more details later.

td.vue/ai-tools/src/main.py
InfosecOTB: You are right - the LiteLLM client can’t recognize the model. The default value is set to an empty string ('') to allow the code to continue and return a LiteLLM error explaining that the model is missing. However, if you think there is a better approach for this, please feel free to correct it.

lreading: After testing, the warning is perfectly clear!  I hadn't considered the exception bubbling when reviewing the code.  Disregard!

InfosecOTB:
Good spot @lreading — I hadn’t noticed that the file name was changed as well, so I adjusted only the patch.

Just to share more from my experience: although it was the most expensive option, I achieved the best results using Anthropic’s Claude Opus (at the time of testing, it was version 4.1), followed by GPT-5 and Grok-4. For testing purposes, I used grok-4-fast-reasoning, which is very cheap and generates relatively good results.

I also ran some tests executing the Threat Generation process multiple times with the same and with different models. The results were interesting: up to a certain number of runs, even the same model added new threats and mitigations to existing ones.

Additionally, I believe you can run the AI tools against a self-hosted model, if it is a decent size. I had good results with deepseek-v3.1-terminus (671B parameters) and qwen3-coder (480B parameters) hosted on Novita. I realize this setup is beyond a typical local computer, but it may still be within reach for some company infrastructures. It is also relatively close to NVIDIA DGX Spark–level capabilities (around 200B parameters), so it’s likely that running such models at home will become feasible soon, especially considering the box costs around $4K, as far as I remember.

td.vue/package.json
```diff
@@ -196,7 +197,7 @@
     "ws": ">=8.18.0",
     "yargs-parser@>=6.0.0 <13.1.2": ">=13.1.2"
   },
-  "main": "background.js",
+  "main": "index.js",
```
lreading: Technically out of scope, but just a small change to address tech debt.  We probably had a `background.js` file some time ago, but now the real entrypoint is index.js after building.  There was a harmless console warning, but this addresses that.

lreading: This is why scope creep is bad.  This actually broke the snap packaging, so I'm reverting.  Bad Leo!

InfosecOTB:
Hi @lreading ,

Since we have run into issues with the integration, and rewriting the code in JavaScript will take some time, I created a user-friendly GUI version of my original program to make the tool available to users in the meantime.

I hope you find it interesting. I would greatly appreciate your feedback, and please let me know if you have any questions or comments.

GitHub repository:
https://github.com/InfosecOTB/threat-dragon-ai-tool 

My blog article about the app.
https://infosecotb.com/ai-powered-threat-modelling-with-owasp-threat-dragon-part-3-threat-dragon-ai-tool/

PR #1418, opened 2025-12-22
Extendable EoP games diagram

javiermorenov1203:
**Summary**:
This is a version of EoP games diagram which we believe is more extendable. 
@jgadsden @sydseter I create this PR in case you prefer to have this version.

A game selector has been added to ThreatEditDialog when the type of the diagram is EoP.
The games are added in td.vue\src\service\threats\models\eop\index.js
Whenever a new game is added, a new file should be created in this directory.
For now, the file should have the same methods as cornucopia.js since these are used in ThreatEditDialog.


javiermorenov1203:
@jgadsden the corresponding modification has been made

jgadsden:
> the corresponding modification has been made

Thst is perfect @javiermorenov1203 , thank you and going ahead with the merge now

PR #1424, opened 2026-01-01
Fix SVG/PNG export to exclude selection decorations

### LINKED ISSUE #1318: When exporting a diagram to a graphics format the selected items are shown selected in the exported image
**Describe the bug**:

The colleague who pointed out #1317 also pointed out another odd behavior regarding the Export feature from the diagram view. It can also be seen in #1317 for the two zoomed views.

When exporting a diagram to an image the selected state gets exported:

Image

This is unexpected insofar as the grid is also shown in the UI but doesn't get exported as visible grid into the images.

Tested with the following export formats:

* PNG: Image
* JPEG: Image (turns out this is in fact a PNG ... will open another ticket for that issue)
* SVG: Image

**Expected behaviour**:

I would expect that the selection state doesn't get exported in any visible manner.

From the user perspective I don't really care if this happens by "remembering" the selection, then deselecting, then generating the image and then re-selecting the previous selection or via another mechanism.

**Environment**:

- Version: 2.4.1
- Platform: Desktop App
- OS: Windows 10 (fully updated)

**To Reproduce**:

* Load [new-model.zip](https://github.com/user-attachments/files/21040919/new-model.zip) from #1317
* Open the one diagram in it
* Select the process item in the middle
* Export into any of the target image formats

* [new-model.zip](https://github.com/user-attachments/files/21040919/new-model.zip) from #1317 was also used to show the behavior for _this_ issue

sn3ha-dev278:
**Summary**:  

Fixes an issue where selected diagram elements were included in exported SVG and PNG files.
Closes #1318.

Remove selection highlights from exported SVG and PNG diagrams.

**Declaration**:  

- [ ] appropriate unit tests have been created / modified
- [ ] functional tests created / modified for changes in functionality
- [x] any [use of AI](../blob/main/contributing.md#use-of-ai) has been declared in this pull request

**Other info**:

### DIFF
--- a/td.vue/src/components/GraphButtons.vue
+++ b/td.vue/src/components/GraphButtons.vue
@@ -150,22 +150,21 @@ export default {
             });
         },
         async withSelectionCleared(fn) {
-            // Store currently selected cells
+            
             const selectedCells = this.graph.getSelectedCells();
             
-            // Clear the selection to remove visual highlights
-            this.graph.cleanSelection();
-            
-            // Wait for next tick and a small delay to ensure rendering completes
-            await this.$nextTick();
-            await new Promise(resolve => setTimeout(resolve, 100));
             
             try {
-                // Perform the export
+                this.graph.cleanSelection();
+
+                //Rendering is not immediate. Without this pause the export may include
+                //the previous selection highlight.
+                await new Promise(resolve => setTimeout(resolve, 100));
+                
                 fn();
             } finally {
-                // Restore the selection after a brief delay
-                await new Promise(resolve => setTimeout(resolve, 50));
+                
+                
                 if (selectedCells.length > 0) {
                     this.graph.select(selectedCells);
                 }


td.vue/src/components/GraphButtons.vue
```diff
@@ -137,15 +137,40 @@ export default {
...
+            // Wait for next tick and a small delay to ensure rendering completes
+            await this.$nextTick();
+            await new Promise(resolve => setTimeout(resolve, 100));
+            
+            try {
+                // Perform the export
+                fn();
+            } finally {
+                // Restore the selection after a brief delay
+                await new Promise(resolve => setTimeout(resolve, 50));
```
jgadsden: time-delay probably not needed, consider removing and retesting

td.vue/src/components/GraphButtons.vue
```diff
@@ -137,15 +137,40 @@ export default {
...
+        },
+        async withSelectionCleared(fn) {
+            // Store currently selected cells
+            const selectedCells = this.graph.getSelectedCells();
+            
+            // Clear the selection to remove visual highlights
+            this.graph.cleanSelection();
+            
+            // Wait for next tick and a small delay to ensure rendering completes
+            await this.$nextTick();
```
jgadsden: again time-delay probably not needed, consider removing and retesting

jgadsden:
tested with and without time-out delays, both work OK and no race condition detected
@sn3ha-dev278 this is an identical solution to one that used AI to provide the solution, can you confirm whether you used generative AI or not?

td.vue/src/components/GraphButtons.vue
sn3ha-dev278: Thanks for the review.

I re-tested this and found that clearing the selection alone isn’t enough—the export can still capture the old selection highlight. The state updates immediately, but the visual update happens asynchronously, so a short delay gives the UI time to finish rendering before export.
I kept the delay, mentioned the reason in the code, and confirmed that both PNG and SVG exports no longer include selection highlights.

Issue #1445, opened 2026-02-03
Bug: navigation error with self-hosted GitLab repo


ia-jordan:
**Describe the bug**:  

When i access the branch of my repo I get an error and it tries to navigate to some object :

```https://<server>:<port>/#/git/gitlab/security%2Fthreatdragon/%5Bobject%20Object%5D/threatmodels```

instead of the real branch. Probably a type error.

in the domain:
"%5Bobject%20Object%5D"

I am using a self hosted gitlab + threatdragon + nginx

**Expected behaviour**:  

Should navigate to the correct branch

**Environment**:  

- Version: owasp/threat-dragon:v2.5.0
  - `gitlab 18.8.2-ce.0`
- Platform: Web App
- OS: Windows with Linux docker
- Browser: Edge

**To Reproduce**:  

1. Configure self hosted GitLab
2. Try to open a model
3. access branch e.g. "main"
4. shows strange %5Bobject%20Object%5D
5. (also wants to navigate to github.com/* if i click in the ui)

**Declaration**:  

By submitting this issue you have:  

- [x] read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [x] not used agentic or generative AI in creating this bug report

**Any additional context, screenshots, etc**:  

also shows this in the ui:
<img width="1866" height="567" alt="Image" src="https://github.com/user-attachments/assets/d35eab83-633d-4f29-af8a-8115996c83b5" />

ia-jordan:
Hi @jgadsden I already tried that but also got the error. I'll attach my setup (Windows + wsl docker), maybe you can reproduce it. (I cleaned up the files to run on localhost instead of my machine name - maybe you need to change that back)

[gitlab-cleaned.zip](https://github.com/user-attachments/files/25363409/gitlab-cleaned.zip)

Ajith05105:
Ahhhh great catch , I’ll make a quick patch to the Dockerfile. I can see that the specific link is being hardcoded to GitHub in the backend, so I’ll work on a quick fix for both of those.

<img width="668" height="210" alt="Image" src="https://github.com/user-attachments/assets/2cbf4069-e7b7-4847-9985-8bf490f5d266" />
    
this is where that url is being built from and its always pointing to github

Thanks for reporting these, and for the great attention to detail :))

PR #1454, opened 2026-02-17
Fix/gitlab error opening models

### LINKED ISSUE #1450: Bug: Opening an existing model from Gitlab returns 500
**Describe the bug**:  

I connected my gitlab selfhosted instance and can see the repositories, yet when you try and open the tm7/json file it fails to load it.
The link also shows some `/` to get encoded (the ones inside subprojects) 
`/mg%2Frd%2Fin%2Fot%2Fgw%2Fdl%2F01%2Fgateway/aggiornamento-loadingpane/threatmodels`

**Expected behaviour**:  

Correct opening of the file.

**Environment**:  

- Version: 2.5.0
- Platform: VM
- OS: RHel
- Browser: chrome

**To Reproduce**:  

- Connect Gitlab following the docs.
- Try opening a project inside a complex tree 

Image

Ajith-Penmatsa-GGL:
**Summary**:  
Fixes two GitLab provider issues:

500 error when ThreatDragonModels directory doesn't exist.  GitBeaker's GitbeakerRequestError doesn't set e.statusCode, so the existing 404 check in the controller didn't catch it. Moved the error handling into gitlabrepo.modelsAsync to return an empty list when the directory is not found, so now its consistent with GitHub provider behavior. This fix closes #1450 .

TypeError: Cannot read properties of undefined (reading 'title') when creating a new threat model from ThreatModelSelect — the newThreatModel() method was duplicating model creation logic and dispatching tmActions.create before setting the model in the store. Replaced with a route to the existing NewThreatModel view which already handles this correctly.(This is something i found while fixing #1450 so no issue was raised for this)

image

So this is the first change where I added a simple try-catch block at the GitLab repository layer since it’s related to GitLab-specific REST API calls. If a 404 is caught here, it returns an empty array back to the frontend, and the frontend handles this gracefully by allowing the user to create a new threat model.

image

ThreatModelSelect.newThreatModel() dispatched tmActions.create before setting the model data in the Vuex store, causing state.data.summary.title to be undefined. Replaced the inline logic with a route to the existing NewThreatModel view which handles the correct sequence.

image

Fix GitLab 500 error on missing ThreatDragonModels directory and fix new model creation from ThreatModelSelect.

**Declaration**:  

  - AI Tools: Claude 
  - LLMs and versions: Sonnet 4.5
  - Prompts: mainly used it to find the mismatch between existing error handling and api response

**Other info**:

### DIFF
--- a/td.vue/tests/unit/views/threatmodelSelect.spec.js
+++ b/td.vue/tests/unit/views/threatmodelSelect.spec.js
@@ -209,27 +209,20 @@ describe('views/ThreatModelSelect.vue', () => {
             getLocalVue({
                 params: {
                     provider: mockStore.state.provider.selected,
-                    repository: mockStore.state.repo.selected
+                    repository: mockStore.state.repo.selected,
+                    branch : mockStore.state.branch.selected
                 }
             });
             wrapper.vm.newThreatModel();
         });
 
-        it('clears the threat model', () => {
-            expect(mockStore.dispatch).toHaveBeenCalledWith(THREATMODEL_CLEAR);
-        });
-
-        it('creates the threat model', () => {
-            expect(mockStore.dispatch).toHaveBeenCalledWith(THREATMODEL_CREATE, expect.anything()); 
-        });
-
-        it('navigates to the edit page', () => {
+        it('navigates to the new threat model page', () => {
             expect(mockRouter.push).toHaveBeenCalledWith({
-                name: 'gitThreatModelEdit',
+                name: `${wrapper.vm.providerType}NewThreatModel`,
                 params: {
                     provider: mockStore.state.provider.selected,
                     repository: mockStore.state.repo.selected,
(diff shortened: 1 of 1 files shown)


Ajith-Penmatsa-GGL:
Actually in theory the existing documentation is correct  GITLAB_REPO_ROOT_DIRECTORY is optional and defaults to ThreatDragonModels. The expected behaviour is:

- If the ThreatDragonModels directory doesn't exist in the selected branch, the frontend prompts the user to create a new model

- On creation, the app automatically creates the ThreatDragonModels directory and saves the model within it

The issue was that GitBeaker's(GitLab's api  library) error format differs from octonode (GitHub's api library) — it doesn't expose statusCode on the error object, so the existing 404 catch in the controller never matched, resulting in a 500 instead of an empty model list.

Just for a sanity check , I'll spin up a container using the Docker Compose from issue #1445 and confirm that this fix works.

PR #1467, opened 2026-03-01
fixed rendering issues with template docs 

Ajith-Penmatsa-GGL:
**Summary**:  

The templates documentation page had several rendering issues. One image had a typo in its path, causing it to fail to load and display alt text instead. Images were also inconsistently sized , some too large, some too small , making the page look unpolished. Some sections also lacked screenshots, leaving users without guidance on how to use the feature.

**Before :**
https://github.com/user-attachments/assets/5a563584-0eb7-44aa-8539-5f15ef831b23

**Changes** 
Fixed the broken image path. Added additional screenshots to improve clarity on how to use the feature.

**After :"

https://github.com/user-attachments/assets/a6ad604d-db8e-435c-945f-2a49e3c0b355

**Declaration**:  

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,0 +1,38 @@
+source "https://rubygems.org"
+# Hello! This is where you manage which Jekyll version is used to run.
+# When you want to use a different version, change it below, save the
+# file and run `bundle install`. Run Jekyll with `bundle exec`, like so:
+#
+#     bundle exec jekyll serve
+#
+# This will help ensure the proper Jekyll version is running.
+# Happy Jekylling!
+gem "jekyll", "~> 4.3.4"
+
+# Theme
+gem "owasp-td-jekyll", "~> 0.2.14"
+
+# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
+# uncomment the line below. To upgrade, run `bundle update github-pages`.
+# gem "github-pages", group: :jekyll_plugins
+# If you have any plugins, put them here!
+group :jekyll_plugins do
+  gem "jekyll-feed", "~> 0.17"
+end
+
+# Windows and JRuby does not include zoneinfo files, so bundle the tzinfo-data gem
+# and associated library.
+platforms :mingw, :x64_mingw, :mswin, :jruby do
+  gem "tzinfo", ">= 1", "< 3"
+  gem "tzinfo-data"
+end
+
+# Performance-booster for watching directories on Windows
+gem "wdm", "~> 0.2.0", :platforms => [:mingw, :x64_mingw, :mswin]
+
+# Lock `http_parser.rb` gem to `v0.6.x` on JRuby builds since newer versions of the gem
+# do not have a Java counterpart.
+gem "http_parser.rb", "~> 0.6.0", :platforms => [:jruby]
+
+# HTTP server toolkit
+gem "webrick", "~> 1.9"
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -1,0 +1,174 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    addressable (2.8.7)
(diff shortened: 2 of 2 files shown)


docs/usage/templates.md
```diff
@@ -1,163 +1,180 @@
...
-
+ 
 Administrators can update a template's name, description, or tags without modifying the template content.
-
-1. Navigate to the Manage Templates
+ 
+1. Navigate to the Manage Templates page
 2. Find the template you want to update
-3. Click the burger menu on the template card and click on edit
+3. Click the burger menu on the template card and select **Edit**
```
jgadsden: some (very obsessive) users may regard this is a kebab-button that leads to the template menu rather than a burger/hamburger button
up to you on how you word it, you could use a generic term 'menu button' but that is less informative than what you have now

docs/usage/templates.md
```diff
@@ -1,163 +1,180 @@
...
 ## Using Templates
-
+ 
 [Threat Dragon](https://owasp.org/www-project-threat-dragon/) supports templates that allow you to
 quickly create new threat models based on pre-defined structures.
-
+ 
 **Note**: The template feature is currently only available for GitHub repositories
 and local users (if they have templates in the local file system).
 Support for Atlassian Bitbucket, GitLab, Google Drive and desktop is not yet implemented.
```
jgadsden: The changes to the docs read very well, thank you @Ajith-Penmatsa-GGL 
Just a suggestion for sentence "Support for Atlassian Bitbucket, GitLab, Google Drive and desktop is not yet implemented." which reads in maybe a negative way.
Did you want to reword it so that it is more upbeat ? along the lines of "will be in place sometime soon"?
Although my rewording does not read very well, so I would not go with that literally :)

docs/usage/templates.md
```diff
@@ -1,163 +1,180 @@
...
+## Loading a template from a local file
+ 
+ 
+In addition to organisation templates, you can also start from a local template file.
+This is useful when you have received a template file from a colleague.
+ 
+**Note**: Template files use a different schema to standard threat model files and are not interchangeable.
+To use an existing model as a starting point, first export it as a template using the **Export as Template** option.
+ 
+From the Template Gallery, click **Import from a Local Template** button to browse for a template file
```
jgadsden: Is the button 'Start from a Local Template' ?

docs/usage/templates.md
Ajith-Penmatsa-GGL: After a quick Google search, I realised you’re right, it’s called a kebab button, not a burger menu, which is the three dashed lines. I’m happy to change it to kebab button since that seems more appropriate. What do you think about calling it “three-dot menu” or “template options button”? I’m not sure how common those terms are, just throwing a few ideas out there.

docs/usage/templates.md
Ajith-Penmatsa-GGL: Ahhh you’re right, great catch. I’ll fix that.

docs/usage/templates.md
Ajith-Penmatsa-GGL: I agree with that I’ll definitely reword it to make it sound more positive.

Ajith-Penmatsa-GGL: The template feature is currently available for GitHub repositories and for local users who have templates stored in their local file system. Support for Atlassian Bitbucket, GitLab, Google Drive, and desktop is coming in future updates. does that sound better?

docs/usage/templates.md
jgadsden: yes, I agree either of your versions sound good and go with your decision, I have a slight preference for “three-dot menu”

docs/usage/templates.md
jgadsden: yes, that sounds more positive, thank you!

jgadsden:
@Ajith-Penmatsa-GGL could we have `Gemfile` and `Gemfile.lock` reinstated?

PR #1473, opened 2026-03-07
Dependency Vuln Cleanup

lreading:
**Summary**:  

Our Trivy action has been failing due to vulnerabilities in dependencies.  This PR addresses those and also does some cleanup/maintenance type work.

Upgraded dependencies to resolve vulnerabilities, updated docker image to use node 24

**Declaration**:  

- [ ] appropriate unit tests have been created and/or modified
  - AI Tools: `OpenAI Codex`
  - LLMs and versions: `GPT-5.3-codex`
  - Prompts:
    - Please find out why I am getting a permission error when building the docker image locally, but not when it runs in CI.
    - Please run the trivy scan locally - trivy is in the path, and we have our trivyignore file in .github/workflows/.trivyignore
    - Please review each CVE that is ignored in .github/workflows/.trivyignore and determine if we can remove them.  If we can, please remove them, build a new image, and run trivy again to prove there are no failures

**Other info**:  

I manually reviewed all changes, most of the real work was done by a human.  Some notes:
- I thought this was a good opportunity to clean up old ignores in `.trivyignore` - gives us good safety from regressions for vulns that were previously fixed
- Found a _local only_ bug when building the docker image.  The translation files generated by cornucopia had permissions issues in the intermediate container that broke the docker build
- Node v20 EOL is next month or so.  We need to update.  I've been using node 24 for local development for a while now with no issues in TD.  Instead of going incrementally, node 24 gives us the longest EOL window and is officially an LTS release
- When running an unrelated npm command, it re-ordered the dev dependencies.  I decided to keep it just to prevent inadvertent package.json updates in future PRs
- Lychee (linkchecker) keeps getting rate-limited by GH.  The [action's documentation](https://github.com/lycheeverse/lychee-action?tab=readme-ov-file#utilising-the-cache-feature) highlights using a cache as a mitigation to rate limiting.  I've updated the action to the latest version and added a cache to see if it helps


.github/workflows/browserstack.yaml
```diff
@@ -20,10 +20,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6.0.2
 
-      - name: Use node LTS 20.14.0
+      - name: Use node LTS 24.14
```
lreading, line 23: In our Dockerfile, we are only specifying the Node major/minor version - we are not pinning to a specific patch version.  Because of that, I think it makes sense to have our actions only specify major/minor as well.  

Our previous strategy was accidentally introducing Node version drift - the docker image was on a later version of node than what we were using in our tests.

Let me know if you think we should pin to a patch version instead. I don't have a strong preference here.  Minor pinning means less maintenance but also a _slightly_ higher risk of breaking changes.  I just ask that we be consistent across our Dockerfile and actions to help avoid future drift. 😎

td.server/package.json
```diff
@@ -17,8 +17,8 @@
     "start:server": "pm2 --name td.server start 'nodemon dev.js --exec babel-node'",
     "stop:server": "pm2 delete td.server",
     "transpile": "babel ./src --out-dir dist",
-    "test": "cross-env NODE_ENV=test nyc mocha --require @babel/register --all true --timeout=900 test/*.spec.js test/**/*.spec.js",
-    "test:unit": "cross-env NODE_ENV=test nyc --reporter=lcov --reporter=text-summary mocha --require @babel/register --all true --timeout=900 test/*.spec.js test/**/*.spec.js"
+    "test": "cross-env NODE_OPTIONS=--no-experimental-detect-module NODE_ENV=test nyc mocha --require @babel/register --all true --timeout=900 test/*.spec.js test/**/*.spec.js",
```
lreading, line 20: https://github.com/OWASP/threat-dragon/pull/1473#issuecomment-4016537658

.github/workflows/.trivyignore
```diff
@@ -1,33 +1,16 @@
-# ignoring these vulnerabilities in zlib,
```
lreading, line 1: This is where the robots helped me most - I was able to test if these were still present.  All of the ones that were removed from this file were no longer present in our image.  Removing them from the ignore file allows us to catch any potential future regressions.

jgadsden: good point @lreading , and as you declared your use of AI _and_ provided the prompt then you are in conformance with our guidelines :)

.github/workflows/browserstack.yaml
jgadsden: Major/Minor is perfectly OK with me - I think the only reason for path is if we wanted to be sure a security patch was included

td.server/package.json
jgadsden: beautiful!

jgadsden (approved):
nice to have all this this corrected and up to a sensible Node version, Thanks @lreading !

Issue #1495, opened 2026-03-15
Translation to Portuguese


jgadsden:
**Describe what problem your feature request solves**:  

At present the translation file `td.vue/src/i18n/pt.js` is in Brazilian Portuguese, we should also have a translation file for Portuguese as it is generally spoken in Portugal

**Describe the solution you'd like**:  

- existing file `td.vue/src/i18n/pt.js` moved to `td.vue/src/i18n/pt-br.js`
- new file `td.vue/src/i18n/pt.js` created for Portuguese
- files `td.vue/src/desktop/menu.js` and `td.vue/src/i18n/index.js` modified to import and display new PT translation
- tests modified for new translation in `td.vue/tests/unit/desktop/menu.spec.js`

**Declaration**:  

By submitting this issue you have:  

- [x] read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [x] not used agentic or generative AI in creating this feature request

**Additional context**:  

after discussion with @AndreFerreiraMsc at the OWASP leaders summit 2026

lreading:
Completed in #1503 - thanks again, @AndreRFerreira !!!

Issue #1502, opened 2026-03-16
Bug: gaps in translation files not filled by EN version


jgadsden:
**Describe the bug**:  

Some translation strings are not being retrieved

**Expected behaviour**:  

if there is a gap in the translation file then the i18n package should retrieve the EN translation string

**Environment**:  

- Version: 2.6.0-latest
- Platform: both Web App and Desktop App
- OS: all of MacOS / Windows / Linux
- Browser: all (chrome / firefox / safari + others)

**To Reproduce**:  

1. navigate to https://www.threatdragon.com/#/dashboard
2. choose language such as Deutsch
3. observe createFromTemplate is not replaced with a translation string
4. open a threat model
5. observe exportTemplate is not replaced with a translation string

**Declaration**:  

By submitting this issue you have:  

- [x] read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [x] not used agentic or generative AI in creating this bug report

**Any additional context, screenshots, etc**:  

<img width="515" height="578" alt="Image" src="https://github.com/user-attachments/assets/0ce9ac38-b13f-4cb9-a781-d5f99ca8e0f3" />
<img width="433" height="234" alt="Image" src="https://github.com/user-attachments/assets/ad9db4c8-cc98-4c96-91b5-4b2c9252bd86" />

AndreRFerreira:
As part of a larger refactor aimed at reducing future maintenance and tackling complexity, I propose the following changes:

- Rename all i18n constants to "messages".
- Rename pt-br.js to pt-BR.js (clarity on BCP 47)
- Update `i18n/index` to use 2 letter country codes in imports. This also allows us to remove the fallbackLocale block for Brazilian portuguese, as VueI18n will handle fallback automatically.
- Update other files that currently use three-letter language codes to use two-letter codes instead.
- Add unit tests that demonstrate:
-- A missing Brazilian Portuguese (BR) key falls back to European Portuguese (PT), and vice versa.
-- If both are missing, it falls back to English
- Keep track of how behind is each language on `docs/development/translations.md` for easier access

I have this half coded, but haven't got around to finish it (Easter break), expect this some time next week.

PR #1503, opened 2026-03-16
Feat/pt-br-translation

AndreRFerreira:
**Summary**:  

Closes [issue https://github.com/OWASP/threat-dragon/issues/1495] 
This PR adds European Portuguese alongside Brazilian Portuguese

Add support for both Brazilian and European Portuguese translations

**Declaration**:  

**Other info**:  

Renames pt.js to pt-br.js (Brazilian Portuguese)
Creates new pt.js with European Portuguese translations
Updates locale selection to support both variants
Updates menu components and tests accordingly
Includes review of technical terminology (e.g., "linkability" → "associabilidade")
Automated tests have been run locally (npm test) and passed.
Introduces fallbackLocale to deal with absent translation keys, mostly between pt-PT and pt-BR but also English

### DIFF
--- a/td.vue/src/i18n/index.js
+++ b/td.vue/src/i18n/index.js
@@ -16,6 +16,7 @@ import ind from './id.js';
 import jpn from './ja.js';
 import msa from './ms.js';
 import por from './pt.js';
+import bra from './pt-br.js';
 // hide RUS & UKR for now: import rus from './ru.js';
 import spa from './es.js';
 // hide RUS & UKR for now: import ukr from './uk.js';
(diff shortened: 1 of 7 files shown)


AndreFerreiraMsc:
Please hold review. I want to ensure that "porBr" is the best option here. Thinking that it may break automatic language detection (if in place), will check later today

td.vue/src/i18n/index.js
```diff
@@ -28,7 +29,12 @@ const get = () => {
     if (i18n === null) {
         i18n = new VueI18n({
             locale: 'eng',
-            messages: { ara, deu, ell, eng, spa, fin, fra, hin, ind, jpn, msa, por, zho }
+            fallbackLocale: {
```
jgadsden, line 32: thanks for sorting this out @AndreFerreiraMsc and correcting bug #1502 
it is possible that the i18n package changed behavior somewhere in the versions so that `locale` was not used as the default and hence the bug

jgadsden:
you make a very good point @AndreFerreiraMsc , and my vote is to go with 'BCP 47 only' alignment
As you say it makes much more sense and we get to using 'pt-BR'

The only problem is that we can not use this as a variable identifier within Javascript code as it means 'pt minus BR', but we can get round that easily :)

The user should never see 'por' or 'pt-PT' for example, as the drop down list is populated with the translated string, so all from good my side

Issue #1518, opened 2026-03-22
Trivy Supply Chain Attack Transparency Report (No Impact)


lreading:
# Threat Dragon Was Not Impacted

AquaSecurity's Trivy GitHub Action was compromised, including an overwrite of 76 existing release tags.
For more information on the technical details of this breach, please see [CrowdStrike's Blog Post](https://www.crowdstrike.com/en-us/blog/from-scanner-to-stealer-inside-the-trivy-action-supply-chain-compromise/) and [Trivy's official statement](https://www.aquasec.com/blog/autonomous-runtime-security-turning-runtime-intelligence-into-agentic-response-2/).  The Actions were compromised on March 19, 2026 at ~17:43 UTC, and containment was completed on March 19, 2026 at ~20:38 UTC

**Threat Dragon was NOT impacted by this compromise**. Even so, our community deserves to know what happened, how we investigated, and what we could do differently to safeguard against future supply chain attacks.  Per our [community trust documentation](https://www.threatdragon.com/docs/trust/trust.html), this issue is being created as a transparency report to our community.

---

## How Threat Dragon Uses Trivy

Threat Dragon uses the official Trivy Action to scan our built docker images for known vulnerabilities:

- on every [pull request](https://github.com/OWASP/threat-dragon/blob/main/.github/workflows/pull_request.yaml#L465) targeting `main`
- on every [push](https://github.com/OWASP/threat-dragon/blob/main/.github/workflows/push.yaml#L491) to `main`
- [daily](https://github.com/OWASP/threat-dragon/blob/main/.github/workflows/housekeeping.yaml#L65)

We have some light documentation on how/why Trivy is used in the [Trust / Container scanning](https://www.threatdragon.com/docs/trust/container.html) and [Trust / Dependency management](https://www.threatdragon.com/docs/trust/dependencies.html) pages.

---

### Trivy runs pre-compromise

_At the time these jobs ran, `aquasecurity/trivy-action@0.34.0` pointed to the legitimate, unmodified code.  The malicious force-push did not occur until `2026-03-19T17:43:00Z`._

| Run ID | Trivy job started (UTC) | Trivy job completed (UTC) | Trigger | Workflow | Outcome |
|---|---|---|---|---|---|
[...]

PR #1516, opened 2026-03-22
Fix Desktop e2e Menu Test to Use Latest Release

lreading:
**Summary**:  

Fixes #1517 
I already broke the desktop e2e tests! :disappointed: 
There's a test that checks all the links in the menu.  In #1497 we updated the menu to point to the latest release.
I reviewed and approved the PR, but I didn't think to update #1494 :sweat_smile: 

**Description for the changelog**:  

Fix for desktop e2e test

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [x] content meets the [license](../blob/main/license.txt) for this project
- [x] appropriate unit tests have been created and/or modified
- [x] you have considered any changes required for the functional tests
- [x] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [x] *either* no AI-generated content has been used in this pull request
- [ ] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie, etc]`
  - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro, etc]`
  - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

**Other info**:
  
Whoops!

If we decide that these tests are too flaky in the future, I'm happy to revisit!  We could even reel it in to use a startsWith type matcher.  I think for now my preference would be to keep this as is, so it is immediately obvious in future PRs if we are changing any of the external links.

PR #1519, opened 2026-03-22
Pinning Trivy GH Action to commit hash for v0.35.0

lreading:
**Summary**:  

Relates to #1518 

Updates all Trivy GH Actions to v0.35.0 - but **using the commit hash instead of a version tag**.

Our Trivy actions are failing because the tags they reference are no longer available.  All tags were removed as part of the containment effort by AcquaSecurity (the makers of Trivy) related to a recent compromise.  

Unlike tags, commit hashes are immutable.  Even if the git history is rewritten, the commit becomes orphaned rather than overwritten.

Pin Trivy GitHub action to specific commit

**Declaration**:  

- [ ] appropriate unit tests have been created and/or modified

**Other info**:  

This is the exact recommendation from AquaSecurity: https://www.aquasec.com/blog/autonomous-runtime-security-turning-runtime-intelligence-into-agentic-response-2/

I think there needs to be a larger discussion around GH Actions / dependency pinning.  Aggressive pinning like this does increase maintenance overhead, and it is also less readable than a tag which is usually a semantic version.  However, due to the immutable nature of commit hashes, this is significantly more secure from a dependency management / supply chain hardening perspective.

### DIFF
--- a/.github/workflows/housekeeping.yaml
+++ b/.github/workflows/housekeeping.yaml
@@ -62,7 +62,8 @@ jobs:
           ref: main
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.34.0
+        # Pinned to commit for v0.35.0
+        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1
         with:
           image-ref: '${{ env.IMAGE_NAME }}'
           format: 'template'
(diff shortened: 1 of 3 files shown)


lreading:
To check for successful credential rotation, I manually ran the CI pipeline: https://github.com/OWASP/threat-dragon/actions/runs/23407258044

- the desktop e2e tests failed (expected, see #1516 )
- Heroku deploy passed (successful rotation)
- Browserstack automation worked (successful rotation)

No other keys were rotated, so I think we are good.

jgadsden (approved):
that seemed to do it, thanks @lreading for the investigation

PR #1542, opened 2026-04-09
Update dependencies in workflows and application

jgadsden:
**Summary**:  

* use `npm-run-all2` which is a supported version of `npm-run-all`
* update vulnerable dependencies identified by Trivy scan
* ignore sites that block lychee automation

**Description for the changelog**:  

update dependencies in workflows and application

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [x] content meets the [license](../blob/main/license.txt) for this project
- N/A appropriate unit tests have been created and/or modified
- N/A you have considered any changes required for the functional tests
- [x] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [x] *either* no AI-generated content has been used in this pull request
- [ ] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie, etc]`
  - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro, etc]`
  - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

**Other info**:

PR #1527, opened 2026-03-31
docs: add documentation for Threat Dragon AI tool (#1493)

InfosecOTB:
**Summary**:  

Add documentation for the Threat Dragon AI Tool.
Fixes #1493

**Description for the changelog**:  

Add documentation for the Threat Dragon AI tool, including a new usage guide and supporting screenshot, and link the new page into the docs navigation.

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [x] content meets the [license](../blob/main/license.txt) for this project
- [ ] appropriate unit tests have been created and/or modified
- [ ] you have considered any changes required for the functional tests
- [x] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [ ] *either* no AI-generated content has been used in this pull request
- [x] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie, etc]`
  - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro, etc]`
  - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`
Cursor IDE was used during the documentation creation process.

PR #1545, opened 2026-04-09
Refresh docs theme for search, nav, accessibility, and bugfixes

### LINKED ISSUE #1544: Docs: Navigation Refresh, Search, Accessibility, and Bug Fixes

When reading the documentation, I found myself using the browser's native search function to quickly find information.  Given I'm a maintainer of Threat Dragon, I'm probably _more familiar_ with the docs than our average user.  This inspired me to add a search feature to the Jekyll template.

I've released a new version of the Jekyll template that includes:
- Search
- Accessibility check / fixes (required in CI now)
- Improved header UI (icons with tooltips over long text that doesn't scale well)
- Updated a vulnerable gem
- Fixed a text overflow bug with in left-nav
- Restore mobile nav / UI improvements
- Fixed collapsed menu behavior
- Removed unnecessary files bundled into the template (README, gemspec, etc)

- Update the `owasp-td-jekyll` gem in docs to [v2.1.0](https://github.com/lreading/owasp-td-jekyll/releases/tag/v2.1.0)
- Enable search
- Convert header links to icon/tooltip
- Update light/dark theme switcher tooltip text 

**Declaration**:  

**Additional context**:  

The following screenshots are from the owasp-td-jekyll repo's test harness.  The PR I open will include screenshots highlighting the changes as they relate to TD's docs.

**Search**

Image

I am actively avoiding adding new third party dependencies to the theme.  Aside from the recent supply chain issues, the "standard" way of leveraging third party css/javascript in Jekyll is loading them as lib/vendor files.  That pattern makes it difficult to identify known vulns and manage versions.

Jekyll creates a JSON file from the content during build, which is consumed by some light javascript.  The feature set of the search is intentionally limited - I didn't want to pull in a heavy third party dependency for "basic search".  This also means that there is no semantic search, or helpers like a Levenshtein distance correction feature.  This is an intentional tradeoff.

**Header Links**

Image

The existing header links contain text and an icon.  There was no tooltip.  I added a configurable tooltip config so that we can use _only the icons_. This scales the navbar better on smaller devices, and made room for the search feature.  

**Sidebar Nav Text Overflow**

_Existing version of the theme, fixed in the new version_:

Image

This was a simple CSS/overflow bug that is fixed in the newest version.

**Mobile Nav**

The mobile nav has been broken for a long time due to a missing viewport.  The existing mobile nav was difficult to use and didn't match the theme well.  I restyled it to make it more usable. This screenshot has multiple dropdowns open just for show.

Image

**Accessibility**

The repo's actions now run `html-validate`, which provides _some_ coverage for accessibility.  While not perfect, I [[addressed multiple findings](https://github.com/lreading/owasp-td-jekyll/pull/40). This check is required to pass for PRs and releases.  In the future, as I learn more about a11y, I hope to expand the scope.

lreading:
**Summary**:  

Closes #1544 

- Update `owasp-td-jekyll` to the [latest release](https://github.com/lreading/owasp-td-jekyll/releases/tag/v2.1.0).
- Search
- "Icon only" header links
- Restore mobile nav
- Fix known vuln in gem dependency
- Fix text overflow bug in sidebar nav
- Improve accessibility

Docs: add search, improve accessibility, restore mobile nav, UX improvements, misc bugfixes

**Declaration**:  

**Other info**:  

Homepage:
image

Search:
image

Side nav overflow fix:
image

Mobile Nav:
_This is a bit goofy looking because of how long the title is, but I don't think it's worth changing.  We don't have stats, but I suspect mobile users are not our target audience anyways_.
image


docs/_config.yml
```diff
@@ -13,18 +13,25 @@ brand: OWASP Threat Dragon Documentation
 faviconUrl: assets/icons/threatdragon.ico
 logo_src: assets/images/threatdragon_logo_image.svg
 
+search:
```
lreading, line 16: Search is configurable.  
Any page with a title is included by default.
You can [exclude individual pages](https://github.com/lreading/owasp-td-jekyll?tab=readme-ov-file#search) by adding `search: false` to the frontmatter.

both `tooltip` and `text` are valid options.  If you use text, it will show the search icon and the text.

docs/_config.yml
```diff
@@ -13,18 +13,25 @@ brand: OWASP Threat Dragon Documentation
...
+  tooltip: Search Threat Dragon documentation
+
+theme_toggle:
+  dark_tooltip: Switch to dark mode
+  light_tooltip: Switch to light mode
+
 header_links:
-  - text: Try the web app
+  - tooltip: Try the web app
```
lreading, line 24: `text` is still a valid option, and would retain the old behavior (text next to icon)

_My preference_ is the tooltip approach.  On medium-sized devices, the navbar quickly runs out of room.  Rather than getting clever with CSS or JS to auto-hide text, I think it makes sense to just lean on tooltips and have it be icon-only.  I like the look better too, but this is just one person's opinion! 😎

lreading:
Leaving in draft until #1542 is sorted.  The gemfile.lock update is likely to cause a conflict, and that work is already in flight.

lreading:
Ready for review!

I did bump axios to keep Trivy happy. The latest release is 3 days old - if that's a bit too young please let me know and we can just fix trivy later!

docs/_config.yml
jgadsden: agreed, the updated look is impressive, and I agree with tooltip instead of text

jgadsden:
There seems to be an issue with the 'Integrations' from previous PR #1527 : 

Screenshot 2026-04-10 at 06 44 17

This is because there is only one page in 'Integrations' (it can be tested by duplicating a page using `nav_order` 1) so 'Integrations' remains highlighted
We could create a general integrations page with nav order 0? and bump `usage/ai-tool.md` to nav order 1?

lreading:
Thanks for finding that!  

1. Your screenshot looks different from what I see when I test. :thinking: Those links that are inactive are really dark and don't have enough contrast. :thinking: Screenshot below - but to be honest I have _no idea_ why this would happen.
2. We absolutely can reorder however we want, but IMO the theme should support top-level pages (nav_order: 1) without us having to work around it. This is definitely a bug, and I'll fix that / push a new version - once working as expected we can still have the conversation about _how_ we want to organize our pages. :smile: 

Here's what my sidebar looks like (the extra highlight is there, but the inactive links are much brighter on my screen):

image

jgadsden (approved):
all good - please merge, the only difference was that I had my contrast turned right down (I am that sort of guy)
looks like most things are in place for a version 2.6.1 release sometime over the weekend?

lreading:
Ahhhh, that explains it!

Is this what you usually see in other dark-mode websites, or is there room for improvement in the theme?  If that's expected, all good! :sunglasses: 

I think it's still worth fixing the nav menu items before pushing, that should be a pretty quick and easy fix.  I'll have this updated with a newer version of the theme soon.

lreading:
Because the drop-downs and links are both "clickable", I think the right solution is to have them both share the same style.  It makes sense, intuitively, to have a different style for something that expands versus something that routes.  In practice, it makes for an inconsistent experience and will likely cause confusion.

Pushed v2.1.1, and updated the gemfile.

image

PR #1597, opened 2026-04-22
Setting least priv permissions for workflows

lreading:
**Summary**:  

CodeQL has correctly warned us that:
> If a GitHub Actions job or workflow has no explicit permissions set, then the repository permissions are used. Repositories created under organizations inherit the organization permissions. The organizations or repositories created before February 2023 have the default permissions set to read-write. Often these permissions do not adhere to the principle of least privilege and can be reduced to read-only, leaving the write permission only to a specific types as issues: write or pull-requests: write.

chore: set GH actions permissions

**Declaration**:  

**Other info**:  

I did my best to reason about each action and what permissions they'd need, but there's no guarantee that I caught everything! :sweat_smile:

### DIFF
--- a/.github/workflows/housekeeping.yaml
+++ b/.github/workflows/housekeeping.yaml
@@ -78,34 +78,6 @@ jobs:
         with:
           sarif_file: 'trivy-results.sarif'
 
-  codeql:
-    name: Analyze with codeql
-    runs-on: ubuntu-24.04
-    permissions:
-      contents: read
-      security-events: write
-    strategy:
-      fail-fast: false
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          ref: main
-
-      - name: Initialize CodeQL
-        uses: github/codeql-action/init@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
-        with:
-          languages: 'javascript'
-          config-file: ./.github/codeql/codeql-config.yml
-          # If you wish to specify custom queries, you can do so here or in a config file.
-          # By default, queries listed here will override any specified in a config file.
-          # Prefix the list here with "+" to use these queries and those in the config file.
-
-      - name: CodeQL autobuild
-        uses: github/codeql-action/autobuild@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
-
-      - name: Perform vulnerability analysis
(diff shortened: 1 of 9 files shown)


lreading:
Ironic :laughing: 
image

.github/workflows/housekeeping.yaml
```diff
@@ -53,7 +56,7 @@ jobs:
     name: Scan with Trivy
     runs-on: ubuntu-24.04
     permissions:
-      contents: write
+      contents: read
```
jgadsden, line 59: This needs `write` permissions because of the tidy workflows and tidy pull-requests actions, but I could be mistaken

.github/workflows/push.yaml
```diff
@@ -480,7 +484,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: build_docker_image
     permissions:
-      contents: write
+      contents: read
```
jgadsden, line 457: just checking that this is the case? that `read` is OK here?

jgadsden (approved):
couple of questions, but easiest to merge this in and keep an eye on any lack of privileges that may arise
nice that CodeQL is earning its keep, or so it seems so :)

lreading:
Instead of guessing, I'll add a temporary workflow dispatch to those jobs and run manually to verify.

.github/workflows/push.yaml
lreading: Without codeql passing, this one is a bit harder to test/validate. 
However, my understanding is that having the `contents: read` is fine so long as we have the `security-events: write` permission (we do below).  I think it's the same as the trivy action which now has a similar permissionset.

lreading:
:facepalm: There were already workflow dispatch triggers for these. :laughing: 

https://github.com/OWASP/threat-dragon/actions/runs/24781528331/job/72513709850
- Removed stale workflow **runs**
- Did not remove stale **workflows** - so that is still untested

As I understand it, we don't need `content: write` in the housekeeping workflow because we set [actions: write](https://github.com/OWASP/threat-dragon/blob/main/.github/workflows/housekeeping.yaml#L17) for the tidy workflows job, and [pr/issues write](https://github.com/OWASP/threat-dragon/blob/main/.github/workflows/housekeeping.yaml#L36) for the tidy prs job.

Unable to test the codeql permissions at the moment (see my other comment), but I'm _reasonably_ confident this should be good to go!

lreading:
Holding off on the merge - it seems I botched the merge conflict and the codeql jobs removed in #1598 are being added back.

.github/workflows/push.yaml
jgadsden: OK, makes sense

PR #1630, opened 2026-05-09
Fix inconsistent state between cell data and name for component labels

lreading:
**Summary**:  

The best I (and my robot minions) can tell is that this is a race condition introduced by the Vue3 compat runtime. I took a TDD approach, creating a failing test first then the fix.

The properties panel used BootstrapVue’s update event for the name field, but the handler ignored the emitted value for the input. Calling dataChanged.updateName(cell) reads cell.getData().name, but that read can happen before v-model has written back to cell.data.name. The result was cell.setName(...) receiving the previous value.

Closes #1627 

**Description for the changelog**:  

bugfix: component labels one keystroke behind

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [x] content meets the [license](../blob/main/license.txt) for this project
- [x] appropriate unit tests have been created and/or modified
- [x] you have considered any changes required for the functional tests
- [x] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [ ] *either* no AI-generated content has been used in this pull request
- [x] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `Codex`
  - LLMs and versions: `GPT5.5`
  - Prompts: `Identify the bug in issue 1627, and create a failing unit test to prove it` `Fix the bug, ensure all tests pass, explain in detail the root cause`

**Other info**:

PR #1633, opened 2026-05-10
add vue key to graph properties to prevent stale cell references

### LINKED ISSUE #1632: Bug: initial confusion for diagram node properties
**Describe the bug**:  

Properties of one diagram component incorrectly displayed for another component. Initially when diagram nodes are placed on the diagram, the properties of one component can be dispalyed for another component

Note that this corrects itself if the component is deselected and then selected again

**Expected behaviour**:  

When selecting a newly placed component on the diagram, the properties are the default and not a previously selected component properties

**Environment**:  

- Version: 2.6.1 only
- Platform: both Web App and Desktop App
- OS: all of MacOS / Windows / Linux
- Browser: all

**To Reproduce**:  

1. create a new threat model with a new diagram
2. edit the diagram and place two components on it (for example processes)
3. edit the name of the selected component
4. select the other component and note that the properties are of the original component, not the newly selected component

**Declaration**:  

- [x] not used agentic or generative AI in creating this bug report

This is a minor bug only, not a priority, as it can be easily corrected / worked round

lreading:
**Summary**:  

Closes #1632 

This seems like a regression from #1630 

The root cause appears to be a stale reference in the GraphProperties.vue for selected cell, which was sometimes using the bootstrap form references instead of the true cell reference.  The fix is to add a cell key so that the _cell_ is always referenced rather than the bootstrap form.  This causes the properties panel to be redrawn when the cell reference changes, and prevents the stale reference trap.

bugfix: fix component properties to always update when selected cell changes

**Declaration**:  

  - AI Tools: `Codex`
  - LLMs and versions: `GPT-5.5`
  - Prompts: `Find the root cause of #1632`

**Other info**:  

This was a combination of human-written code and AI assisted coding.  Codex correctly identified the root cause, but its code suggestions were terrible, so I rewrote it by hand.  The unit test fails prior to the cell key fix, and they pass after.  I manually tested as well.

### DIFF
--- a/td.vue/src/components/GraphProperties.vue
+++ b/td.vue/src/components/GraphProperties.vue
@@ -41,7 +41,10 @@
       </b-col>
     </b-row>
 
-    <b-form v-if="!!cellRef && cellRef.data">
+    <b-form
+      v-if="!!cellRef && cellRef.data"
+      :key="cellKey"
+    >
       <b-form-row>
         <b-col md="6">
           <b-form-group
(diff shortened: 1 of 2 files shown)


jgadsden (approved):
tested on both macOS and Linux and the bug has been fixed
superb to get the fix in so quickly, thanks @lreading

PR #1674, opened 2026-05-28
Supply Chain: enforce minimum release age

### LINKED ISSUE #1673: Supply Chain: enforce minimum release age

Supply chain attacks are becoming more common than ever. There are multiple ongoing campaigns, and staying on the latest version has become a big security risk.

Our Dependabot config enforces a minimum release age of 10 days for everything [except security updates](https://github.com/dependabot/dependabot-core/issues/14683).

There's some light research that studied 10 recent supply chain attacks and found that a [14 day cooldown](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns#fn:apt) would have prevented 9 out of 10.

This is currently an _undocumented convention_, and not enforced anywhere outside of the dependabot config and maintainer process.

- Align on an acceptable minimum release age 
- Update the [dependency docs](https://www.threatdragon.com/docs/trust/dependencies.html) to include the agreed upon minimum release age
- Add `.npmrc` to the root, td.vue, and td.server directories
  - I'd recommend doing one in all 3, even via symlink if better. My concern is that developers may open only the `td.vue` directory to work, and that _may_ not honor the root `.npmrc`
  - include `min-release-age=x` where x is the agreed upon minimum release age
- update `Dockerfile` to ensure `.npmrc` is copied before installing dependencies (might be overkill where it's `npm ci`, but better safe than sorry)
- audit actions to ensure `.npmrc` is honored

**Declaration**:  

**Additional context**:  

I've been happy enough with 10 days, but I'm really not married to it. If we'd like it longer, that's perfectly fine by me.

Related: #1669

lreading:
**Summary**:  

Closes #1673 

- Add / update `.npmrc` files to include a minimum release age of 10 days for root, td.vue, and td.server
- Remove hard-coded `--legacy-peer-deps` in favor of `.npmrc`
- Requires npm version >= 11.10.0 to ensure the minimum release age flag works
- Update docs to include the minimum release age setting and npm version floor
- Update Dockerfile to use `.npmrc` files, including the `npm i -g npm@latest`, this was one of the remaining supply-chain gaps

chore: enforce a minimum release age of 10 days for all npm dependencies

**Declaration**:  

**Other info**:  

There is another supply chain gap that I want to address, but I prefer having a separate issue/PR for it: we are relying on tags for our Dockerfile base image. Tags are mutable, and Dockerfile supports and recommends pinning by sha256. Dependabot supports the Docker ecosystem. 

The reason I am leaving that out of scope is for node version drift. We would have to be careful to update all of the NODE_VERSION flags every time a Dockerfile base image is updated. I also want to make sure that the cooldown works as expected, _and_ that it supports the version comments the same way we manage github actions.  This could be automated as well.

I opted out of a `.node-version` implementation for now, but am happy to reconsider. My concern is that if we pin the node-version too hard, it becomes more difficult to contribute. I think enforcing a strict node version for the build and deploys makes sense, but we are already doing that. I don't think pinning the node version adds enough value to justify forcing node version management into the developer experience. (feel free to disagree!)

I also did not use `engine-strict=true` because td.vue depends on `@achrinza/node-ipc@9.2.2`, which only declares node version support through v17 :woozy_face: The engine-strict flag causes the install to fail. Hopefully we can revisit that after the vue3 migration is complete.

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -256,8 +256,7 @@ jobs:
             ${{ runner.os }}-
 
       - name: Install clean packages
-        # TODO: remove --legacy-peer-deps after the Vue 3 migration
-        run: npm clean-install --legacy-peer-deps
+        run: npm clean-install
 
       - name: Build Linux desktop app
         run: npm run build:desktop -- --linux AppImage --x64 --publish never
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -295,8 +295,7 @@ jobs:
             ${{ runner.os }}-
 
       - name: Install clean packages
-        # TODO: remove --legacy-peer-deps after the Vue 3 migration
-        run: npm clean-install --legacy-peer-deps
+        run: npm clean-install
 
       - name: Build Linux desktop app
         run: npm run build:desktop -- --linux AppImage --x64 --publish never
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 !.markdownlint.yaml
 !.wordlist.txt
 !.spellcheck.yaml
+!.npmrc
 !CNAME
 
 # docker files
@@ -108,6 +109,7 @@
 !td.server/test/repositories/
 !td.server/test/repositories/*.spec.js
 !td.server/.babelrc
+!td.server/.npmrc
 !td.server/.eslintrc.js
 !td.server/README.md
 !td.server/dev.js
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+min-release-age=10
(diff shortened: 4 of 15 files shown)


lreading:
@jgadsden - I want to make sure we're aligned on the minimum release age before moving forward.

Is 10 days sufficient, or would you prefer something longer? Happy to update to whatever you think is best!

jgadsden (approved):
thanks @lreading , good to have this in place

jgadsden:
> Is 10 days sufficient, or would you prefer something longer?

I missed this question first time round, and from my point of view 10 days is a good amount. I have seen 3 days used elsewhere which for an open source project is probably too short, so 10 days gives time to block / respond

lreading:
Appreciate your thoughts! I had taken the approval as alignment, my bad! :laughing: 
We can always revisit this too - I'd say let's see how things pan out.

PR #1691, opened 2026-06-04
Pin Docker base images

lreading:
**Summary**:  

Pins the docker base images and adds a dependabot config to manage them.
Closes #1675 

The original issue was suggesting a lot of automation and moving parts. It felt a bit too fragile.
Where I added a couple of maintenance scripts including one to upgrade the node version across actions, I think it's unnecessary. 

**Description for the changelog**:  

chore: pin docker base images

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [x] content meets the [license](../blob/main/license.txt) for this project
- [x] appropriate unit tests have been created and/or modified
- [x] you have considered any changes required for the functional tests
- [x] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [x] *either* no AI-generated content has been used in this pull request
- [ ] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie, etc]`
  - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro, etc]`
  - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

**Other info**:

PR #1692, opened 2026-06-04
Optimize Dockerfile

boomer41:
**Summary**:  

The current Dockerfile produces a rather huge image (~500 MB incl. node, ~340MB excl. node) and has a lot of layers.
Additionally, there are many dependencies in that Dockerfile that ensure that the build cache can't be used with the most efficiency.

This PR optimizes the Docker Image and pulls it down to 223 MB incl. node and 64MB excl. node.

Additionally, I enabled a multiplatform build for arm64 in addition to amd64.
This would produce one image with multiple platforms, making the second -arm64 image obsolete.

- Reworked Dockerfile to reduce image size and enhance cache usage
- Switched to use `@googleapis/drive` instead of global `googleapis` packages to reduce size
- Removed separate `-arm64` image and enabled multiplatform builds for amd64 and arm64
- Enabled SLSA provenance and Docker SBOMs for the images

**Declaration**:  

- [X] content meets the [license](../blob/main/license.txt) for this project
- [X] appropriate unit tests have been created and/or modified
- [X] you have considered any changes required for the functional tests
- [X] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [X] *either* no AI-generated content has been used in this pull request

**Other info**:  

Might need a rebase on https://github.com/OWASP/threat-dragon/pull/1691 after that is merged.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -150,15 +150,6 @@ jobs:
         with:
           install: true
 
-      - name: Cache Docker layers
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ hashFiles('Dockerfile') }}
-          restore-keys: |
-            ${{ runner.os }}-buildx-
-            ${{ runner.os }}-
-
       - name: Login to Docker Hub
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
         with:
@@ -174,8 +165,8 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           push: true
           tags:  ${{ env.IMAGE_NAME }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache,mode=max
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
           # If you add more platforms, make sure to include them in ./pull_request.yaml too!
           platforms: linux/amd64,linux/arm64
           # Do not load the image heare, as multiplatform images can't be loaded that way
@@ -196,13 +187,6 @@ jobs:
           path: './boms/*'
           include-hidden-files: true
           if-no-files-found: error
-
-      - # Temp fix for large cache bug
-        # https://github.com/docker/build-push-action/issues/252
-        name: Move cache
-        run: |
-          rm -rf /tmp/.buildx-cache
-          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
   heroku_deploy:
     name: Upload to Heroku
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -165,9 +165,8 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-      # platform manifests not (yet) supported, so split out architectures
-      - name: Build  for amd64 and push latest
-        id: docker_build_amd64
+      - name: Build for supported architectures and push latest
+        id: docker_build
         uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
         with:
           context: ./
@@ -177,22 +176,11 @@ jobs:
           tags:  ${{ env.IMAGE_NAME }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache,mode=max
-          platforms: linux/amd64
-          load: true
-
-      - name: Build for arm64 and push latest-arm64
-        id: docker_build_arm64
-        uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
-        with:
-          context: ./
-          file: ./Dockerfile
-          builder: ${{ steps.buildx.outputs.name }}
-          push: true
-          tags:  ${{ env.IMAGE_NAME }}-arm64
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
-          platforms: linux/arm64
-          load: true
+          # If you add more platforms, make sure to include them in ./pull_request.yaml too!
+          platforms: linux/amd64,linux/arm64
+          # Do not load the image heare, as multiplatform images can't be loaded that way
+          # The following docker run will instead pull from the registry we just pushed to
+          load: false
 
       - name: fetch app SBOMs
         run: docker run --rm --entrypoint tar "$IMAGE_ID" -c boms | tar -xv
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -173,7 +173,7 @@ jobs:
           # The following docker run will instead pull from the registry we just pushed to
           load: false
           sbom: true
-          provenance: mode=max
+          provenance: mode=max,version=v1
 
       - name: fetch app SBOMs
         run: docker run --rm --entrypoint tar "$IMAGE_ID" -c boms | tar -xv
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -167,9 +167,9 @@ jobs:
           tags:  ${{ env.IMAGE_NAME }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          # If you add more platforms, make sure to include them in ./pull_request.yaml too!
+          # If you add more platforms, make sure to include them in ./pull_request.yaml and ./release.yaml too!
           platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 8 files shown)


boomer41:
I have pushed a test image built with this PR under 

`docker.io/boomer41/owasp-threat-dragon-reverse-proxy-test:latest@sha256:c6b5d38039c5c8795fbee033f842b57351b6157f622ea01d4605ef421a037799`

It was built with:

`docker buildx build . --push --platform=linux/amd64,linux/arm64 -t boomer41/owasp-threat-dragon-reverse-proxy-test:latest
`

boomer41:
What is the intended way for the trivy failures? Should I patch the two mentioned versions up, or is this done via dependabot?

lreading:
> What is the intended way for the trivy failures? Should I patch the two mentioned versions up, or is this done via dependabot?

This is actually coming from the npm version in the final image. 
In the `npm-build-base` image - `npm i -g npm@latest`, but the final image does not. 
As of today, that means that the builder image uses npm@11.15.0, but the final image is using npm@11.13.0 which is what has the vulnerable transient dependencies.

Dockerfile
```diff
@@ -1,64 +1,106 @@
...
+FROM build-npm-base AS build-npm-stage1
+
+COPY package-lock.json package.json ./
+
+COPY ./td.server/.npmrc ./td.server/package-lock.json ./td.server/package.json ./td.server/
+
+COPY ./td.vue/.npmrc ./td.vue/package-lock.json ./td.vue/package.json ./td.vue/
+COPY ./td.vue/src/plugins ./td.vue/src/plugins
+
+RUN npm install --ignore-scripts && \
```
lreading: `npm install` will overwrite the package-lock.json files and potentially bring in new versions of dependencies. `npm clean-install` (or ci) preserves and honors the lockfile.

Thinking something like this instead?
```
RUN npm clean-install --ignore-scripts && \
    cd td.server && npm clean-install --ignore-scripts && \
    cd ../td.vue && npm clean-install --ignore-scripts
```

Dockerfile
```diff
@@ -1,64 +1,106 @@
...
+    cd td.server && npm install && cd .. && \
+    cd td.vue && npm install
+
+# For some reason, the build also wants Electron...
```
lreading, line 31: Yeaaahhhhh, this is weird. 😅 

We have a vue plugin `vue-cli-plugin-electron-builder`. During build, vue loads all plugins (both dev and prod). Under the hood, the electron builder plugin eagerly loads electron, then electron will try to download the binary on load.

I toyed around with a couple ideas to get around this before, but both were more complex than they were worth (at least to me!).

Dockerfile
```diff
@@ -1,64 +1,106 @@
...
+
+# NPM: Stage 2: install runtime-only dependencies for the server
+# This stage is kept as minimal as possible to ensure aggressive caching
+# Should only be rebuilt when the package.json or package-lock.json change,
+# and not the TD source code itself
+FROM build-npm-base AS build-npm-stage2
+
+COPY package-lock.json package.json ./
+COPY ./td.server/.npmrc ./td.server/package-lock.json ./td.server/package.json ./td.server/
+RUN cd td.server && npm install --omit=dev --ignore-scripts
```
lreading: Same as above - `npm clean-install` is preferred to preserve the lockfile. :)

lreading:
Great stuff, @boomer41! Thanks!!!
My only concerns around npm clean-install vs npm install, and the npm version mismatch/trivy.

I closed the dockerfile pinning PR (https://github.com/OWASP/threat-dragon/pull/1691) in favor of this one. I will add the dependabot config after this gets merged. :sunglasses:

boomer41:
I switched to install, because Clean-Install won't do anything new. The node_modules shouldn't exist, but if you want I can ofc switch to clean install.

I don't install npm in the final image, because, well, there is no need, so makes the image smaller :) I'll check if there is a node image without npm. I think the -slim images are more appropriate for the final image. I'll check and update the PR.

jgadsden:
thanks for this @boomer41 
Could we consider using `clean-install` as this should be deterministic?
whereas `install` may pull in dependencies that have been recently compromised through some supply-chain attack for example?

.github/workflows/push.yaml
```diff
@@ -177,7 +177,7 @@ jobs:
           tags:  ${{ env.IMAGE_NAME }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache,mode=max
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
```
jgadsden, line 171: we use linux/arm64 at line 183, so we do not need it for this step?

Dockerfile
jgadsden: the `--ignore-scripts` was added because we used `npm clean-install` at the top level
there are post-install scripts at this level that change to both `td.vue` and `td.server` and run `npm install`, whoch we had to ignore if we want `npm clean-install` in both of these directories

.github/workflows/push.yaml
boomer41: The idea is to make one singular multi-platform image, and not two different. The second dedicated -arm64 below could then be deprecated.
Has also the advantage that the layers for the TD stuff will be re-used for both architectures.

Dockerfile
boomer41: I will change it back to clean-install.

boomer41:
@jgadsden I have updated the Dockerfile according to your comments.
New version is in `docker.io/boomer41/owasp-threat-dragon-reverse-proxy-test:latest@sha256:fb9fce387fcdab4131f8f8f50544d9981b9df1cfc2a525d774e9bccde8b5f781`.

.github/workflows/push.yaml
jgadsden: that is a good point, thanks @boomer41 
could you go ahead and remove the arm64 specific step? then we can go with multi-platform image from now on

boomer41: Removed the separate arm64 stuff. I also enabled building the arm64 image for the tests in PRs and rebased on latest main.

boomer41: I also enabled Docker's dedicated SBOM scanner and enabled max provenance mode.
Provenance includes a detailed description of where the image was built with which exact dependencies and environment. This helps security scanners like Wiz understand the image better.

boomer41:
So. The CI works again. Sorry for the spam, my normal environment is GitLab and CI is notoriously hard to work with locally :)

I had to add a separate amd64-only step for the PR, because multiplatform images are a bit special to handle in the local docker daemon. This isn't problematic, because it will just re-use the already-built stuff from a step above.

Additionally, I switched the pipeline to GitHub Action's Cache instead of a local folder.
Makes it a bit nicer to work with, and a bit more efficient.

I do seriously hope that the release pipeline on main will work, but currently I don't see anything that might fail it.

boomer41:
I also just noticed that there is yet a third place to build a docker image, which I have now modified. Could you please check that I have now caught every place? If something goes wrong with a release, feel free to ping me and I'll take a look. :)

lreading:
This is all excellent, @boomer41! Seriously, **thank you** for the effort you've put into this. On top of the stated goal of reducing the image size, you've also increased trust and the ability to verify Threat Dragon's release artifacts. :heart: I didn't know about Docker's SBOM/provenance mode, I appreciate you giving me the opportunity to learn. In today's world of supply chain attacks, provenance and attestations are becoming increasingly important.

All looks good to me!

---

You found some common friction points - I'm going to share my thoughts on how we can improve developer experience and reduce friction.

> I also just noticed that there is yet a third place to build a docker image, which I have now modified. Could you please check that I have now caught every place? If something goes wrong with a release, feel free to ping me and I'll take a look. :)

We have a _lot_ of repeated jobs across our workflows. GitHub Actions has the concept of reusable workflows - the docker build step is a great use-case. We could define the step in one place and reuse it in the different actions (push, pr, release). If it passes in PR, there's a high level of confidence it will pass during the release workflow. 

I don't expect any issues in the release pipeline, but I can trigger a alpha/pre-release after this merges to verify!

> So. The CI works again. Sorry for the spam, my normal environment is GitLab and CI is notoriously hard to work with locally :)

Not spam at all - we get it, and I think we've all played the CI dance! :wink: :sob: 
If it's helpful, [Act](https://github.com/nektos/act) exists to run actions locally, but I've had mixed luck with it.

I recently added dev/maintenance [scripts](https://github.com/OWASP/threat-dragon/tree/main/scripts). I'll add a new script that emulates the PR workflow to make local debugging easier.

lreading (approved):

boomer41:
No problem :) I also suggest always looking at the final image with something like dive. This reveals what is actually baked in the image, as the old one for example had a lot of npm caches in the image too.

jgadsden (approved):
very good to have, many thanks @boomer41 
do you want to go ahead and merge ? The honor is yours

boomer41:
Well, of course something did not work :) I'll look at it and send a new PR

lreading:
If it helps, we have a workflow dispatch trigger on the action that failed, which just means we can trigger it against any branch we want at any time. :) 

https://github.com/OWASP/threat-dragon/actions/workflows/push.yaml

image

I'm not sure how permissions work there though, let me know if that would be helpful and/or if you're able to trigger it!

PR #1696, opened 2026-06-05
add accepted, transferred and avoided threat statuses

### LINKED ISSUE #1242: Add accepted, transfered, avoided status for threats
For each threat, we can select status "N/A", "Open", "Mitigated". 
According to most standards/publications, risk treatment options include accept, mitigate, avoid, transfer.

Add 3 more options for threat status:
- accepted
- transferred
- avoided

**Additional context**:

Transferred and accepted are mentioned in the three-tier demo model (postgres process), but only as free text in the mitigations field, while the status is "mitigated".

samiat4911:
**Summary**:  
Right now a threat can only be N/A, Open or Mitigated. This adds three more statuses, Accepted, Transferred and Avoided, so the standard risk treatment options can be recorded directly instead of being written into the mitigation text (the three-tier demo model currently fakes "transferred"/"accepted" as free text while leaving the status as Mitigated).

The three new statuses are treated as resolved: they don't mark a diagram element as having open threats, they're hidden by the "show mitigated" toggle in the report, and they're counted as resolved in the report summaries. I moved the status values and the open/resolved logic into one small helper so they're defined in a single place, and mapped the new statuses to the nearest TM-BOM control status on export.

closes #1242

Add Accepted, Transferred and Avoided risk treatment statuses for threats

**Declaration**:  

  - AI Tools: `Claude Code`
  - LLMs and versions: `Claude Opus 4.8`
  - Prompts: `I used Claude Code to implement the change. I picked and scoped the issue, chose the status names (Accepted/Transferred/Avoided) and the "treat as resolved" behaviour, then reviewed the diff and ran the tests before submitting. The model wrote the implementation and unit tests.`

### DIFF
--- a/td.vue/src/assets/schema/threat-dragon-v2.schema.json
+++ b/td.vue/src/assets/schema/threat-dragon-v2.schema.json
@@ -358,7 +358,7 @@
                             "type": "string"
                           },
                           "status": {
-                            "description": "The threat status as NA, Open, Mitigated, Accepted, Transferred or Avoided",
+                            "description": "The threat status as NA, Open, Mitigated, Accepted, Transferred, Avoided or Eliminated",
                             "type": "string"
                           },
                           "threatId": {
--- a/td.vue/src/assets/schema/threat-dragon-v2.schema.json
+++ b/td.vue/src/assets/schema/threat-dragon-v2.schema.json
@@ -358,7 +358,7 @@
                             "type": "string"
                           },
                           "status": {
-                            "description": "The threat status as NA, Open or Mitigated",
+                            "description": "The threat status as NA, Open, Mitigated, Accepted, Transferred or Avoided",
                             "type": "string"
                           },
                           "threatId": {
--- a/td.vue/src/components/ThreatStatusSelector.vue
+++ b/td.vue/src/components/ThreatStatusSelector.vue
@@ -1,0 +1,93 @@
+<template>
+    <div :id="id" class="btn-group td-threat-status-selector">
+        <td-form-radio-group
+            :id="`${id}-primary`"
+            :value="value"
+            :options="primaryOptions"
+            buttons
+            @input="selectStatus"
+        ></td-form-radio-group>
+        <td-dropdown
+            class="td-treatment-dropdown"
+            :class="{ active: treatmentSelected }"
+            variant="secondary"
+            :text="selectedTreatmentText"
+        >
+            <template #default="{ close }">
+                <button
+                    v-for="status in treatmentOptions"
+                    :key="status.value"
+                    type="button"
+                    class="td-dropdown-item"
+                    @click="selectStatus(status.value); close()"
+                >
+                    {{ status.text }}
+                </button>
+            </template>
+        </td-dropdown>
+    </div>
+</template>
+
+<script>
+import {
+    getPrimaryStatusOptions,
+    getTreatmentStatusOptions,
+    isResolved
+} from '@/service/threats/status.js';
+import TdDropdown from '@/components/Dropdown.vue';
+import TdFormRadioGroup from '@/components/FormRadioGroup.vue';
+
+export default {
+    name: 'TdThreatStatusSelector',
+    components: {
+        TdDropdown,
+        TdFormRadioGroup
+    },
+    model: {
+        prop: 'value',
+        event: 'input'
+    },
+    props: {
+        id: {
+            type: String,
+            default: 'status'
+        },
+        value: {
+            type: String,
+            default: ''
+        }
+    },
+    computed: {
+        primaryOptions() {
+            return getPrimaryStatusOptions((key) => this.$t(key));
+        },
+        treatmentOptions() {
+            return getTreatmentStatusOptions((key) => this.$t(key));
+        },
+        treatmentSelected() {
+            return isResolved(this.value);
+        },
+        selectedTreatmentText() {
+            return this.treatmentOptions.find((status) => status.value === this.value)?.text
+                || this.treatmentOptions[0].text;
+        }
+    },
+    methods: {
+        selectStatus(status) {
+            this.$emit('input', status);
+            this.$emit('change', status);
+        }
+    }
+};
+</script>
+
+<style lang="scss" scoped>
+.td-threat-status-selector :deep(.bv-no-focus-ring) {
+    display: inline-flex;
+}
+
+.td-treatment-dropdown.active :deep(.td-dropdown-toggle) {
+    background-color: #5a6268;
+    border-color: #545b62;
+}
+</style>
--- a/td.vue/src/i18n/en.js
+++ b/td.vue/src/i18n/en.js
@@ -471,7 +471,10 @@ const messages = {
         status: {
             notApplicable: 'N/A',
             open: 'Open',
-            mitigated: 'Mitigated'
+            mitigated: 'Mitigated',
+            accepted: 'Accepted',
+            transferred: 'Transferred',
+            avoided: 'Avoided'
         },
         severity: {
             tbd: 'TBD',
@@ -502,6 +505,9 @@ const messages = {
         threatStats: {
             total: 'Total Threats',
             mitigated: 'Total Mitigated',
+            accepted: 'Total Accepted',
+            transferred: 'Total Transferred',
+            avoided: 'Total Avoided',
             notApplicable: 'Total Not Applicable',
             notMitigated: 'Total Open',
             openCritical: 'Open / Critical Severity',
--- a/td.vue/src/i18n/en.js
+++ b/td.vue/src/i18n/en.js
@@ -474,7 +474,8 @@ const messages = {
             mitigated: 'Mitigated',
             accepted: 'Accepted',
             transferred: 'Transferred',
-            avoided: 'Avoided'
+            avoided: 'Avoided',
+            eliminated: 'Eliminated'
         },
         severity: {
             tbd: 'TBD',
@@ -508,6 +509,7 @@ const messages = {
             accepted: 'Total Accepted',
             transferred: 'Total Transferred',
             avoided: 'Total Avoided',
+            eliminated: 'Total Eliminated',
             notApplicable: 'Total Not Applicable',
             notMitigated: 'Total Open',
             openCritical: 'Open / Critical Severity',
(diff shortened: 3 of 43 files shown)


lreading (changes requested):
Hey @samiat4911 - thanks for taking this on! :tada: 

Based on the discussion in the linked issue, I think we should align on a couple things with @jgadsden before merging:
1. Are we happy scoping this to MATA as opposed to TAME?
2. Jon had suggested updating the label to "Remediations" - is that still desired and in-scope for this PR?

My opinions:
1. I don't hold strong opinions on MATA vs TAME - I think this is a great addition and we can add more later
2. I recommend not changing the label. A label-only change (not schema level) might not be considered a breaking change. However, changing the name of a field without a major version bump does blur that line a bit for me. I could be convinced either way, but I do think it should be out of scope for this PR. :sunglasses: 

---

The code looks good! There's an issue in the UI though - having these extra options overflows the severity input box. This is in full screen view, so if someone is using the desktop client non-maximized, it might be even worse.  Note the "Transferred" button:

image


I think the right solution is converting status to a dropdown.

**We are in the process of removing BootstrapVue components** - please use the custom [TdDropdown component](https://github.com/OWASP/threat-dragon/blob/main/td.vue/src/components/Dropdown.vue).

I tested it locally, and it's a pretty straightforward change. :smile: Here's the gist:

```html
<template>
                          <td-dropdown
                              id="status"
                              variant="secondary"
                              :text="selectedStatusText"
                          >
                              <template #default="{ close }">
                                  <button
                                      v-for="status in statuses"
                                      :key="status.value"
                                      type="button"
                                      class="td-dropdown-item"
                                      @click="threat.status = status.value; close()"
                                  >
                                      {{ status.text }}
                                  </button>
                              </template>
                          </td-dropdown>
</template>

<script>
import TdDropdown from '@/components/Dropdown.vue';
export default {
  computed: {
     selectedStatusText() {
      return this.statuses.find((status) => status.value === this.threat.status)?.text || '';
    }
  }
}
</script>
```

samiat4911:
> Hey @samiat4911 - thanks for taking this on! 🎉
> 
> Based on the discussion in the linked issue, I think we should align on a couple things with @jgadsden before merging:
> 
> 1. Are we happy scoping this to MATA as opposed to TAME?
> 2. Jon had suggested updating the label to "Remediations" - is that still desired and in-scope for this PR?
> 
> My opinions:
> 
> 1. I don't hold strong opinions on MATA vs TAME - I think this is a great addition and we can add more later
> 2. I recommend not changing the label. A label-only change (not schema level) might not be considered a breaking change. However, changing the name of a field without a major version bump does blur that line a bit for me. I could be convinced either way, but I do think it should be out of scope for this PR. 😎
> 
> The code looks good! There's an issue in the UI though - having these extra options overflows the severity input box. This is in full screen view, so if someone is using the desktop client non-maximized, it might be even worse. Note the "Transferred" button:
> 


thanks for the feedback, happy to add more  :)

samiat4911:
@jgadsden what do you think do you have any addition?

samiat4911:
@lreading I've converted the status control to TdDropdown in both the edit and suggest dialogs, which fixes the overflow.

jgadsden:
Thanks @samiat4911 for this pull request and also for raising the initial issue, it is good to get this in place

This issue + change is a larger than it might seem at first, in that we should support both TAME and MATA, without changing the experience too much for the established Threat Dragon community and without breaking existing threat models

The card for threats does not look correct to me:
Screenshot 2026-06-09 at 07 37 01

What I would like to have is the existing radio buttons 'N/A', 'Open' and 'Mitigated' as before
The change needs to come in the form of 'Mitigated' now being a drop down list with:

* Mitigated (default)
* Accepted
* Transferred
* Avoided
* Eliminated

Note the addition of Eliminated, so that we support TAME alongside MATA

jgadsden:
The documentation needs to change to describe the changes to this feature, if that could be put in place as well @samiat4911 ?
(note that our guidance for AI-usage requires that this is not generated by AI)

td.vue/src/i18n/en.js
```diff
@@ -471,7 +471,10 @@ const messages = {
         status: {
             notApplicable: 'N/A',
             open: 'Open',
-            mitigated: 'Mitigated'
```
jgadsden, line 474: ideally these changes should be copied over into the other translation files
this `en.js` file is the default for the strings, but it is good to keep the files in step

td.vue/src/service/threats/status.js
```diff
@@ -0,0 +1,68 @@
+/**
```
jgadsden: has this comment/description been generated by AI ?
It does not seem informative

td.vue/src/service/threats/status.js
```diff
@@ -0,0 +1,68 @@
...
+ * Single source of truth for threat status values and their semantics.
+ *
+ * Stored `value` strings are persisted into saved threat-model JSON, so they are
+ * stable identifiers and must not be renamed. The human-readable labels come
+ * from i18n (see STATUS_TRANSLATION_KEYS) and can be changed freely.
+ *
+ * "Resolved" statuses are those that represent a decided risk-treatment outcome
+ * (the threat is no longer open): Mitigated, Accepted, Transferred, Avoided.
+ */
+export const THREAT_STATUS = {
```
jgadsden: the rest of the Threat Dragon code does not use uppercase for constants, so this needs to be changed to be consistent

td.vue/tests/unit/service/threats/index.spec.js
```diff
@@ -177,6 +177,18 @@ describe('service/threats/index.js', () => {
             expect(threats.hasOpenThreats(data))
                 .toEqual(true);
         });
```
jgadsden, line 229: our `AGENTS.md` asks that 

- All new code must have at least 95% coverage for statements, branches, functions, and lines
- Existing code that's updated should be brought to the same coverage threshold

coud we have `td.vue/tests/unit/service/threats/index.spec.js` brought up to this standard?

td.vue/tests/unit/components/threatSuggestDialog.spec.js
```diff
@@ -0,0 +1,49 @@
+import { BootstrapVue } from 'bootstrap-vue';
```
jgadsden: this is reporting:

```
File                              | % Stmts | % Branch | % Funcs | % Lines |
ThreatSuggestDialog.vue           |   44.27 |    44.89 |   61.66 |    45.6 |
```

jgadsden (changes requested):
many thanks for taking this on @samiat4911 , it really is a much larger change than at first sight because it is changing one of the foundations of the threat models
There are various comments, see how you get on and please ask for clarification / help if needed

td.vue/src/assets/schema/threat-dragon-v2.schema.json
```diff
@@ -358,7 +358,7 @@
                             "type": "string"
                           },
                           "status": {
-                            "description": "The threat status as NA, Open or Mitigated",
+                            "description": "The threat status as NA, Open, Mitigated, Accepted, Transferred or Avoided",
```
jgadsden: Could we also have Eliminated here as well as Avoided

samiat4911:
> many thanks for taking this on @samiat4911 , it really is a much larger change than at first sight because it is changing one of the foundations of the threat models There are various comments, see how you get on and please ask for clarification / help if needed

Sure no problem :)

samiat4911:
@jgadsden so I made some update regarding your feedback. please check it out when you are free thanks

samiat4911:
Hello @jgadsden still waiting on your feedback

td.vue/src/components/ThreatStatusSelector.vue
```diff
@@ -0,0 +1,93 @@
...
+};
+</script>
+
+<style lang="scss" scoped>
+.td-threat-status-selector :deep(.bv-no-focus-ring) {
+    display: inline-flex;
+}
+
+.td-treatment-dropdown.active :deep(.td-dropdown-toggle) {
+    background-color: #5a6268;
```
lreading, line 90: This doesn't quite match the visual style of the form-group buttons when they are selected. Could we please update this to be:
```css
.td-treatment-dropdown.active :deep(.td-dropdown-toggle) {
    background-color: darken($gray, 10%);
    border-color: darken($gray, 12.5%);
}
```

Image

The "darken" approach is what Bootstrap was doing under the hood with the native bootstrap components. We're in an awkward in-between state with the vue/bootstrap migration - these edge cases are very non-obvious. 😅  

In the future I hope to have a proper styleguide, easier to use components that handle active states out of the box, and better global variables for easier consumption!

lreading:
Hey @samiat4911 - apologies for the delay!
This is looking great!!! I added one more comment about the active style for the dropdown active state style. In fairness, that should be handled by another component, but we're just in this weird in-between state with the bootstrap migration at the moment. :sob: 

Everything else is looking great, and again - I genuinely appreciate all the work you've put into this! :tada:

samiat4911:
> Hey @samiat4911 - apologies for the delay! This is looking great!!! I added one more comment about the active style for the dropdown active state style. In fairness, that should be handled by another component, but we're just in this weird in-between state with the bootstrap migration at the moment. 😭
> 
> Everything else is looking great, and again - I genuinely appreciate all the work you've put into this! 🎉

Thanks for the feedback and Goodluck with the migration 😄.

samiat4911:
@lreading can I still contribute to other issues? Or I'll have to wait until the migration is been resolved.

lreading:
@samiat4911 - Yes!!! Absolutely!!! 

I apologize if I gave the wrong impression!  My intent was just to provide more context on why these unusual nitpicks are coming up, and why some of the changes might feel a bit awkward right now. :smile:  I know you've been very active with Threat Dragon, and we really appreciate it. :tada:  I **am not** asking you to slow down or stop by any means.

samiat4911:
> @samiat4911 - Yes!!! Absolutely!!!
> 
> I apologize if I gave the wrong impression! My intent was just to provide more context on why these unusual nitpicks are coming up, and why some of the changes might feel a bit awkward right now. 😄 I know you've been very active with Threat Dragon, and we really appreciate it. 🎉 I **am not** asking you to slow down or stop by any means.

oh alright thanks for the clarity :)

samiat4911:
@lreading would you need my help with #1732 and #1733?

lreading (approved):

lreading:
Hey @samiat4911 - again, sorry for the delay here. I'm gonna go ahead and merge this, and will follow up with a quick PR to handle the little style discrepancy. :)

PR #1706, opened 2026-06-06
Add local script to emulate CI

### LINKED ISSUE #1699: Add script to emulate PR workflow locally

When opening a pull request, our [GitHub action](https://github.com/OWASP/threat-dragon/blob/main/.github/workflows/pull_request.yaml) checks spelling, lints markdown, runs full unit/e2e tests, etc. This is a great way to ensure quality before merging to main.

As with many CI workflows, the developer experience is awkward. Some of these jobs are difficult to run locally without installing additional software and spending time understanding what each step does.

- Create a new [script](https://github.com/OWASP/threat-dragon/tree/main/scripts) that emulates the PR workflow
  - Use bash and keep dependencies to a minimum - docker can be used for anything not in the npm ecosystem
  - Developers on Windows can either use WSL, or I'm sure an AI could easily convert the bash -> batch/powershell
- Update AGENTS.md to use the script if the user is debugging CI
- Update contributing.md to point to the script as a debugging tool

**Declaration**:  

**Additional context**:  

I've experienced this friction personally, which was part of the motivation to start developing the maintenance/local scripts.
The friction also became evident in #1692 for a newer contributor, especially one who is used to a different CI/source control system.

lreading:
**Summary**:  

Closes #1699 

Add a best-effort local script to emulate what runs in CI.
This is helpful when debugging CI failures - running locally takes less time, and _can help_ reduce debugging commits like 'test ci' or 'fix ci'.

chore: add local ci script

**Declaration**:  

  - AI Tools: `Codex`
  - LLMs and versions: `GPT-5.5`
  - Prompts: `Create a local script that runs an approximation of the CI action. Run specific npm commands, and you are permitted to use docker for things that run outside of the node ecosystem. Do not include the e2e:visual tests as they are not consistent across environments`

**Other info**:  

Output: 

image


scripts/td-pr-check.sh
```diff
@@ -0,0 +1,383 @@
...
+            echo "  Note: headless Linux systems may need a display setup outside this script."
+        fi
+    done
+}
+
+need_docker() {
+    docker info >/dev/null 2>&1
+}
+
+desktop_e2e_build_args() {
```
lreading, line 412: running `npm run build:desktop` by itself will attempt to build for all platforms and all arch's. On many systems, the build fails because of missing dependencies and other incompatibilities. 

Where it's only a local dev script, I didn't go nuts hardening this - but it should keep the desktop build scoped to only an app image for the current OS/arch, which should make running the desktop e2es a bit more consistent.

contributing.md
```diff
@@ -86,6 +86,8 @@ The best chance of success for a pull request to be reviewed and accepted:
 7. avoid coding using [generative AI](#use-of-ai-for-coding);
   if you must then understand fully the changes and keep sprawl to a minimum
 
+If CI fails,`scripts/td-pr-check.sh` runs the local equivalents of the PR workflow locally.
```
jgadsden: could this be reworded? it does not quite read right to me

scripts/README.md
jgadsden, line 5: could this intro text be revised to state which directory the scripts should be run from?
When running from bash I found it has to be run from the top / root directory using command `scripts/td-pr-check.sh`
running using `./td-pr-check.sh` within the `scripts` directory breaks because it can not find the config files (of course)
The reason that it is a bit confusing is that developers may be reading this as text rather than in markdown, and see the `./td-pr-check.sh`

scripts/td-pr-check.sh
```diff
@@ -0,0 +1,383 @@
...
+    (
+        cd td.vue || exit
+        npm clean-install &&
+            npm run lint:desktop &&
+            npm run test:desktop
+    )
+}
+
+e2e_smokes() {
+    (
```
jgadsden: this fails on me within bash : 

```
> threat-dragon@2.6.2 test:e2e-pr-smokes
> vue-cli-service test:e2e -C e2e.smokes.ci.config.js --browser chromium --headless --url http://localhost:8080/

Downloading Electron binary...
Error: EEXIST: file already exists, symlink 'Versions/Current/Electron Framework' -> '/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Electron Framework'
    at async Object.symlink (node:internal/fs/promises:1002:10)
    at async Extractor.extractEntry (/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/extract-zip/index.js:130:7)
    at async ZipFile.<anonymous> (/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/extract-zip/index.js:65:11)
/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/electron/index.js:43
      throw new Error(
      ^

Error: Electron failed to install correctly. Please delete `node_modules/electron` and run "npx install-electron --no" manually.
    at getElectronPath (/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/electron/index.js:43:13)
    at Object.<anonymous> (/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/electron/index.js:52:18)
    at Module._compile (node:internal/modules/cjs/loader:1854:14)
    at Object..js (node:internal/modules/cjs/loader:1985:10)
    at Module.load (node:internal/modules/cjs/loader:1577:32)
    at Module._load (node:internal/modules/cjs/loader:1379:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1600:12)
    at require (node:internal/modules/helpers:153:16)
    at Object.<anonymous> (/Users/jon.gadsden/github/threat-dragon/td.vue/node_modules/vue-cli-plugin-electron-builder/lib/testWithSpectron.js:2:22)

Node.js v24.16.0
```

jgadsden:
I like this script a lot, thanks @lreading 
running from this branch I get results  (log files provided if needed) : 

```
Summary:
✓ markdown_lint
✓ link_check
✓ spell_check
✓ server_unit_tests
✓ site_unit_tests
✓ desktop_unit_tests
✗ e2e_smokes
✗ desktop_e2e_smokes
✗ e2e_tests
✓ zap_scan
✓ docker_build
✓ trivy_scan
```

and I am sure it runs OK on your platform but the wider point is that this script may run differently on various platforms?

contributing.md
lreading: "not quite right" is what my mother used to call me.... :trollface: 
Agreed, I took another stab at it! Let me know if the new version reads better! :crossed_fingers:

scripts/README.md
lreading: Great callout, thank you! Updated. :sunglasses: 
I also added a guard clause to validate the PWD (best-effort) to prevent similar confusion even if the developer does not read the readme.

image

scripts/td-pr-check.sh
lreading: Oh Electron... I'm going to need to think about this a bit. :thinking: 

I removed the e2e_smokes check, kind of redundant if we're running the full e2e suite. 
I think it makes sense to keep both `test:e2e` and `test:desktop` though.

It would be nice to have `--skip-checks <comma_separated_check_name>` and `--include-checks <comma_seaparted_check_names>`. On failure, in addition to guidance on running individual checks, it should give an exact command to run using the same script: `./scripts/td-pr-check.sh --include-checks e2e_tests,desktop_e2e_tests`.  I'll update the help text to reflect, and include a listing of all checks by name.

lreading:
> and I am sure it runs OK on your platform but the wider point is that this script may run differently on various platforms?

***Yes*** - emphatically yes. :smile: I will eventually test on windows / WSL once we're happy it runs cleanly on mac/linux. I don't intend to write or maintain batch and powershell equivalents, but am open to community contributions!

I appreciate your help testing as always! I'll try to figure out a better pattern for the e2e tests. These have been pretty fragile - first the visual regression tests, and now the "regular" e2e's. 
I imagine we're doing _something_ that's too prescriptive or opinionated causing the inconsistency. I will dig into it, hopefully it's something obvious! :crossed_fingers:

contributing.md
```diff
@@ -83,8 +83,10 @@ The best chance of success for a pull request to be reviewed and accepted:
 4. provide enough description to explain how Threat Dragon worked before and works after the change
 5. extend the unit tests and functional tests to show the change working as expected
 6. the pull request must provide one change only, if there are further changes use further pull requests
-7. avoid coding using [generative AI](#use-of-ai-for-coding);
-  if you must then understand fully the changes and keep sprawl to a minimum
+7. if coding using [generative AI](#use-of-ai-for-coding);
```
lreading, line 86: @jgadsden - This is out of scope, but something I noticed while editing.

The use of ai for coding section says:
> The use of tools that help write code is encouraged 

This was contradicting that. 
If you disagree, feel it warrants more discussion, or just don't want the scope creep - I'm happy to revert. 😄

jgadsden: good point and good to have this updated 👍🏾

scripts/README.md
jgadsden: just thought that the scripts rely on Docker running on the local machine
did you want to notify the developer about this? Happy either way, it sort of should be obvious

jgadsden:
> I will eventually test on windows / WSL once we're happy it runs cleanly on mac/linux. I don't intend to write or maintain batch and powershell equivalents, but am open to community contributions!

we could write in the readme that these scripts are best effort and YMMV ?
I guess we can assume the contributor is reasonable technical and can work out why it may have broken?

scripts/README.md
lreading: Couldn't agree more! I'll add a dependency guard, and make the contract explicit in the README. 
Scripts should require bash, node/npm, and docker.

Other dependencies should be avoided.

lreading:
I couldn't reproduce the electron install issue on macos, but there is another issue that I've run into repeatedly. Cypress has a browser that's available via NPM - and makes it the theoretical best choice for running tests locally because the dependency lives in `package.json` and not the operating system (like chrome/chromium does).

However, Cypress is an electron wrapper. Our desktop code doesn't make the distinction between running in electron vs running as the desktop application. The default e2e config uses Cypress as a browser. The tests are written for the web app, and there are some nuances between desktop and web causing the tests to fail.

I think we should address this at some point by creating a clear distinction between running as a desktop app and running inside electron - but I don't want to do that here. It's really stretch the scope.

I will update the `test:e2e` to use chrome headless for the interim. There's no guarantee that all developers will have chrome available, but this would be a temporary measure. 

I'm still working through some macOS specific issues, I'll take this out of draft after I have that ready and address the other comments!

lreading:
There were a good handful of cross-platform compatibility issues that were 100% my fault. :disappointed: 

I'll leave some comments on specific fixes and why they matter, but the TLDR is that I really need to start running tests on macos when I write em! :sweat_smile: I don't know how much longer I'm going to have access to a mac - however, I think we could probably use a MacOS runner in a GH action to ensure they pass on both platforms? This would be ideal for the desktop e2e tests anyways, as well as a windows runner. :thinking: I'm not sure if I should continue the scope creep here or not though. :sweat_smile: 

There's a bit more cleanup to do - I added some env var overrides for special use-cases on linux when running tests, so there needs to be documentation updates as well.  Still not quite ready for review, but making solid progress!

td.vue/tests/e2e/desktop/specs/menu.spec.js
```diff
@@ -46,7 +56,7 @@ describe('Desktop menu integration tests', function () {
         const exportTd = getSubmenuItem(exportMenu, fileMenu.labels.exportTd);
 
         assert.equal(fileSnapshot.label, fileMenu.labels.heading);
-        assert.deepEqual(fileSnapshot.submenu.map((item) => item.label), fileMenu.expectedItems);
```
lreading, line 49: This wouldn't ever pass on macOS - macOS adds a "recent documents" item to the menu.
I'll eventually want to test this on windows, but the ordered subset approach seems to work well

td.vue/tests/e2e/desktop/specs/smokes.spec.js
```diff
@@ -16,11 +16,13 @@ describe('Desktop smoke tests', function () {
 
         const windowSize = await browser.execute(() => ({
             width: Math.max(window.outerWidth, window.innerWidth),
-            height: Math.max(window.outerHeight, window.innerHeight)
+            height: Math.max(window.outerHeight, window.innerHeight),
```
lreading, line 19: This was pretty brittle - it was not behaving on macos, and even acts up a bit on Linux if using a tiling WM. I'll get into that more in the config / env overrides.

td.vue/tests/e2e/desktop/support/setup.js
```diff
@@ -102,6 +102,44 @@ const getElectronBinaryPath = () => {
...
+        return;
+    }
+
+    await mainProcessClient.evaluate(`
+        (() => {
+            const electron = process.mainModule.require('electron');
+            const [window] = electron.BrowserWindow.getAllWindows();
+
+            if (window) {
+                window.setResizable(false);
```
lreading, line 141: This is kind of specific to tiling window managers, and I only tested on hyprland. I don't know for sure if this will work on i3 or other WMs. I think this is "good enough" for now, and we can revisit if others are having issues.

td.vue/tests/e2e/desktop/support/setup.js
```diff
@@ -154,9 +192,9 @@ exports.mochaHooks = {
                     '--no-sandbox',
                     '--disable-gpu',
                     '--disable-dev-shm-usage',
-                    '--ozone-platform=x11',
+                    getLinuxOzoneFlag(),
```
lreading, line 203: Many DE's are moving to wayland. The ozone flag matters a lot with electron - without the right flag it might fail to load entirely. 

On macOS, it's entirely unnecessary - which is why we return early in `getLinuxOzoneFlag()` if the platform is not linux.

td.vue/tests/e2e/specs/threatmodel/vue3-regressions.cy.js
```diff
@@ -33,7 +33,8 @@ const openDemoModel = (modelName) => {
     cy.get('#local-login-btn').click();
     cy.get('a[href="#/demo/select"]').click();
     cy.get(`[data-model-name="${modelName}"]`).click();
-    cy.contains(modelName);
+    cy.url().should('contain', `/local/${encodeURIComponent(modelName)}`);
+    cy.get('#td-edit-btn').should('be.visible');
```
lreading, line 37: Flaky test - there are multiple pages that contained the model name. By waiting for the url to load and checking for the edit button, we are technically not testing the model name anymore, but we are testing that the URL is correct and that the page loaded (hence edit button visible). 

Before this fix, I ran the test 20 times and had 2 failures. After - 20 times with 0 failures.

We could add the `cy.contains(modelName)` back if anyone feels strongly that it was providing value. 🤷

td.vue/package.json
```diff
@@ -29,7 +29,7 @@
     "pretest:unit": "node ./src/plugins/generate-cornucopia.js",
     "test": "vue-cli-service test:unit",
     "test:desktop": "vue-cli-service test:unit --testMatch '**/tests/unit/desktop/**/*.spec.js' --collectCoverageFrom 'src/desktop/**/*.js'",
-    "test:e2e": "vue-cli-service test:e2e -C e2e.local.config.js",
+    "test:e2e": "vue-cli-service test:e2e -C e2e.local.config.js --headless --browser chrome",
```
lreading, line 32: This is temporary - I want to eventually standardize on headless cypress as a browser. 
We cannot do that today, because our desktop code only checks for electron env vars, and cypress runs in electron. This means that the front-end will run like a desktop app in the cypress browser no matter what.

I'm not a huge fan of adding external dependencies for developers, but I think it's safe enough for now especially since this is run in CI.

td.vue/vue.config.js
```diff
@@ -86,10 +86,11 @@ module.exports = {
                 },
                 mac: {
                     category: 'public.app-category.developer-tools',
+                    executableName: 'Threat Dragon',
                     icon: './src/icons/icon.icns',
                     hardenedRuntime: true,
-                    entitlements: './node_modules/electron-builder-notarize/entitlements.mac.inherit.plist',
-                    entitlementsInherit: './node_modules/electron-builder-notarize/entitlements.mac.inherit.plist',
+                    entitlements: './node_modules/app-builder-lib/templates/entitlements.mac.plist',
```
lreading, line 91: These changes seemed to fix the desktop tests on macOS, but to be 100% transparent I don't fully understand why.  It could be from all the dependency updates we've been doing recently with the vue3 migration? Something else?

I would love someone else to test on a mac and let me know if things are still working as expected. I'd hate to add another "works on my machine" type of issue.

lreading:
Finally in a better spot. This is working on my linux and mac machines.

PR #1716, opened 2026-06-09
fix(dependabot): remove devEngines.packageManager from package.json

boomer41:
**Summary**:  

See https://github.com/nodejs/corepack/issues/729 for a similar issue.
Dependabot breaks when the package.json specifies a version range for the package manager instead of a specific pinned version.
As the Dockerfile for example already uses the latest npm, this should be redundant.

If you look here, Dependabot does not work:

- Overview: https://github.com/OWASP/threat-dragon/network/updates
- Latest run: https://github.com/OWASP/threat-dragon/network/updates/1400550406

**Description for the changelog**:  

- Fixed Dependabot for npm dependencies

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [X] content meets the [license](../blob/main/license.txt) for this project
- [X] appropriate unit tests have been created and/or modified
- [X] you have considered any changes required for the functional tests
- [X] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [X] *either* no AI-generated content has been used in this pull request
- [ ] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `[e.g. GitHub CoPilot, ChatGPT, JetBrains Junie, etc]`
  - LLMs and versions: `[e.g. GPT-4.1, Claude Haiku 4.5, Gemini 2.5 Pro, etc]`
  - Prompts: `[Summarize the key prompts or instructions given to the AI tools]`

**Other info**:

PR #1727, opened 2026-06-10
Feat/frontend locale config

### LINKED ISSUE #1606: Personalized welcome experience - language selection

When Threat Dragon starts, the initial greeting is displayed in English. Users must manually switch to their preferred language. This step can be eliminated by automatically detecting the user’s browser language via navigator.language, thus respecting their pre defined preferences. The manual language selection would remain available as an override.

Additionally, for organisations or deployments that need to restrict language options (e.g., to support only a subset of languages for policy, legal, or usability reasons), there is currently no way to configure which languages are offered to end users.

I propose two enhancements:

**1. Automatic language detection**
Threat Dragon should default to `navigator.language` instead of hardcoding 'en'. This provides a personalized, localized experience immediately upon launch, while still allowing users to manually change the language later.

**2. Configurable language allowlist (server-side)**
Introduce a new server configuration option that defines which languages are available to users. This will enable administrators to restrict the language selection to a specific subset — for example, only languages officially supported by their organization. While this may intentionally make supported languages unavailable, it ensures consistency for all users in that deployment.

As an added bonus (perhaps): if a particular language translation becomes very outdated, rather than having the allowed list empty by default (or optionally not sent), it can be populated with those available, thus restricting access to it via configuration rather than having to manually comment it out in the code.

**Declaration**:  

- [ x] read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [ ] not used agentic or generative AI in creating this feature request

**Additional context**:  

The idea came to me while having a look under the hood regarding the upgrade of the i18n library. 
I've asked AI to "improve" the feature request text, then I improved, it's "improvement" (hope it works!)

AndreRFerreira:
**Summary**:  
Add configuration-driven locale resolution to support server-published locale policies, locale fallback chains, and restricted locale lists for multi-language deployments. Closes #1606

**Description for the changelog:**
Configuration-driven locale resolution: client reads locale policy from server configuration, resolves user locale against allowed locales with fallback chain, and surfaces configuration load errors in the UI

**Declaration:**

[x] content meets the license for this project
[x] appropriate unit tests have been created and/or modified
[x] you have considered any changes required for the functional tests
[x] you have read the contribution guide and agree to the Code of Conduct
[] either no AI-generated content has been used in this pull request
[x ] or any use of AI in this pull request has been disclosed below:
AI Tools: DeepSeek
LLMs and versions: DeepSeek
Prompts: [Many]

**Other info:**
- Locale resolver (src/service/locale/locale-resolver.js): Pure-function service that resolves user locale against a server-defined allowed-locale list with i18next-style fallback chain. Handles browser navigator.languages, server defaultLocale, and SUPPORTED_LOCALES intersection.
- HTTP client timeout (src/service/api/api.js): Adds configurable default request timeout to prevent hung requests blocking locale resolution (spinners stay forever).
- Config store module (src/store/modules/config.js): Sanitizes server config (validates booleans/strings against SUPPORTED_LOCALES), exposes allowedLocales/defaultLocale getters, and surfaces configError state for error handling.
- Locale store module (src/store/modules/locale.js): Wires the resolver into Vuex — resolves locale on RESOLVE_LOCALE dispatch, selects locale on LOCALE_SELECTED, syncs i18n instance, and enforces the allowed-locale guard.
- HomePage UI (src/views/HomePage.vue): Adds loading spinner while config is fetched, shows toast on config load failure, and extracts resolveProviders() as a pure function for testability.
- LocaleSelect component: Wires locale resolution into dropdown selection.
- i18n constants (src/i18n/index.js): Extracts DEFAULT_LOCALE, SUPPORTED_LOCALES, and installLegacyCompat() for reuse across modules.
- Config load error translation (src/i18n/*.js): Adds home.errors.configLoadFailed key to all locale files.
- E2E tests: Config API mock server, HomePage config-driven tests, locale section tests.
- Unit tests: Full coverage for locale resolver, config store, locale store, config actions, locale actions, and i18n constants.


AndreRFerreira:
Unrelated test **fix coming up** next: resetDesktopApp state pollution in desktop E2E menu tests

**Problem**: The 'opens a new model editor through File > New Model' test was failing intermittently with URL app://./index.html#/desktop/Saved%20As%20From%20Menu/edit instead of the expected app://./index.html#/desktop/New%20Threat%20Model/edit. The URL contained the title from a previous test (SAVED_AS_MODEL_TITLE), indicating cross-test state leakage.

**Root cause**: resetDesktopApp() cleared sessionStorage before browser.refresh(). Between the clear and the refresh, window.location.hash = '#/' triggered Vue Router navigation, causing Vuex mutations to fire. The vuex-persist plugin then re-wrote the stale threat model state back to sessionStorage. When the page reloaded, Vuex re-hydrated from the re-persisted stale data.

**The Fix**: Clear session/localStorage after the first page refresh, then do a second refresh so the app reinitializes from truly empty storage. (td.vue/tests/e2e/desktop/support/helpers.js)

Not related to my feature work — this is a pre-existing test isolation bug in the desktop E2E test infrastructure.

AndreRFerreira:
@lreading I only have chromium, will need assistance to get it to pass the visual test.
Edit: I'll install Chrome later and give it a try

AndreRFerreira:
No luck with the visual test, locally, they are passing but claiming no changes to commit

lreading:
Sorry for the delay @AndreRFerreira - just catching up!

The visual regressions are definitely flaky and don't work well cross-platform. I have a PR in draft that addresses some of that.  I also think I have fixed a couple other flaky tests in #1706 though I'm not sure it fixes the state pollution issue. I'm hoping to get that across the line later today! 

At some point, I want to get all e2e tests back into the Cypress browser. That gives better cross-platform compatibility, but it currently does not work due to our desktop environment detection. The code assumes if it's running in an Electron environment, then it must be the desktop version. Cypress is built on top of electron, so the web e2e tests don't work as intended. :sob: 

I'm going to fix the merge conflicts and take care of the visual regressions! :sunglasses:

lreading:
The visual regressions are failing _now_ for 2 reasons:
1. The landing page has different providers available
2. There's a toast notification showing "Failed to load server configuration..."

For visual regressions, do we even want a backend running? The way I think about visual regressions is a way to identify accidental breaking changes for the most commonly used UI elements as opposed to traditional e2e tests.  

It seems that this toast comes up even if there is no backend running. I don't know if we want that, it could feel spammy to users who run Threat Dragon without a backend. (less common, but technically a supported deployment method).  if there was a td.server configured and the request failed, that failure info would be available in the console/network/server logs. :thinking: 

To keep it concise:
1. What's your take on visual regressions having a backend available? (genuine question!  I have an open mind!:smile: )
2. Is there a way we can avoid showing the toast if there's no backend at all?

---

For reference, here's the home screen visual diff:
 
image

The rest of the diffs all mostly look like this:

image

The slight changes in text location and whatnot are expected, but the toast is what's causing them to hard fail. :smile:

AndreRFerreira:
Thanks for the support @lreading I've never used this before, so I’ve to lean on you here as I'll have to read about it rather than just using.
On the spinner (and both questions), sounds about right, I’ll refactor today, probably add/use a "isElectron" flag, then the issue will go away. I missed that.
Please do let me know on the review of other pieces where I’ve missed anything or for potential improvements.
EDIT: I see a conflict with .gitignore i'll address that later today also
EDIT: I see the issue with the e2e, adding a fixes for this also, hope it's the ones you'r are aware of

AndreRFerreira:
I've taken some liberties in applying other bits and pieces, sorry for the larger review requirement

lreading:
Hey @AndreRFerreira!

Sorry, I had quite a few things come up all at once and needed to take a step back for a couple weeks. 

This is looking great!  I really appreciate all the work you put into it. <3 

I already held this up long enough, so I hope you don't mind me adding a few commits based on things I found during review/testing! :sunglasses: 

1. Because `en` is the default, a _manually_ selected `en` wouldn't persist. For example, if your browser was set to French and you selected English, reloading the page would load it in French again.
2. I made the 5 second axios timeout specific to the config call. I'm hesitant to make that a global config - some providers can be hosted locally, and I don't know that we'd want to cancel requests after 5 seconds (even if 5 seconds is very generous)
3. For now, I've rolled back the no-config toast. Some users may be using Threat Dragon without a backend, and it can be jarring or alarming to have a new error show up. This is technically a supported configuration. I'm open to discuss it, but I think to keep things moving that can be a different discussion. 
4. Visual regression tests don't technically depend on e2e tests, so I think having them only depend on the unit tests will help keep the CI just a bit quicker
5. Tweaked a test assertion and a couple very very minor doc things
6. Unfortunately, dependabot doesn't play nicely with a semver range for devEngines. :sob: See #1716 



Let me know if you disagree with any of the above! Happy to discuss more, of if you'd like to merge I think we're good! :tada:

lreading (approved):

lreading:
I'm gonna go ahead and merge just to avoid any additional code rot from my own delays. If there's any concerns, feel free to comment here, reach out on slack, open issues, smoke signals - whatever works. :laughing:

AndreRFerreira:
Hey @lreading 

1- Nasty miss! Thank you
2- Addressing connectivity and preparation for future enhancement robustness of server communication (in the way I understand enterprise) was the goal, but it was really a suggestion/proposal. I've no reservations on the call/decision, can always be added.
3- Uncertain here, on the Desktop ( IIRC ) it wasn't supposed to show any errors, wondering what you found (if you recall), nevertheless, no worries.
4,5- Good stuff
6- Nice to know, cheers.

Notes and thoughts:
1-
```
this.providerType === providerTypes.desktop VS isDesktopApp()
```
I find the second more readable, but apart from the preference, I would argue that environment is not state given its immutability, however I yield, if there are use cases where that env may change. You will find however, I'll be happy to follow the project preference.

2- On [this doc change](https://github.com/OWASP/threat-dragon/pull/1727/changes/13693e80e0cff57be467a6967709fdf5a7553508)
```
- **Note:** The static fallback is **one‑way** (`pt` → `pt-BR`). `pt` does not fall back to Brazilian Portuguese by default,
+ **Note:** The static fallback is **one‑way** (`pt` → `pt-BR`). `pt-BR` does not fall back to European Portuguese (`pt`) by default,
```
The 3rd party I18N will fall back from `pt-BR` to `pt` automagically (as per the way they coded it).
"`pt-BR` does not fall back to European Portuguese (`pt`) by default" asserts false, "pt" requires manual configuration to pt-BR as the 1st fallback if desired.

3- Love this change :100: , congrats guys
```
Threat Dragon is an [OWASP Lab Project]
Threat Dragon is an [OWASP Production Project]
```

PR #1731, opened 2026-06-13
add OAUTH_FRONTEND_RETURN_URL config and shared oauth helper

### LINKED ISSUE #1539: Configurable OAuth Frontend URL Override / NODE_ENV Documentation

I was following the [GitHub configuration docs](https://www.threatdragon.com/docs/configure/github.html) which outlines `NODE_ENV=development` as a valid configuration.  This configuration uses a hard-coded `localhost` OAuth redirect without an available override.  This is _fine_ for local dev, but it breaks the OAuth integration if deployed elsewhere. When following the docs, a user should be able to stand up the Threat Dragon server without any additional investigation or unexpected behavior.  

The core problem is that `NODE_ENV` has a some smaller, hidden, side-effects.  Those should be explicitly defined in the docs, and we should provide intuitive config-driven overrides.

`NODE_ENV=production`: enables Express rate limiting, disabled otherwise. OAuth redirects are relative urls.

`NODE_ENV=development`: OAuth redirect is hard-coded to `http://localhost:8080/#/oauth-return?...` across all providers

`NODE_ENV=test` only silences logging (not an issue IMO).

The redirect flow for `development` hard-codes not only localhost, but also port 8080.  If using a non-default frontend port locally, you would need to update these hard-coded values in the server.

***Documentation Updates***
- Explicitly define `NODE_ENV` side-effects in the high-level [Configuration Guide](https://www.threatdragon.com/docs/configure/configure.html)
- Update all provider config docs to call out `NODE_ENV` behavior/defaults
- Add `OAUTH_FRONTEND_RETURN_URL` to the high-level config and provider specific docs

***New Configuration Value*** **`OAUTH_FRONTEND_RETURN_URL`**:
- Optional env var with no default
- When not set, follow the current default behavior based on `NODE_ENV`
- When set, redirect to `OAUTH_FRONTEND_RETURN_URL`
- Apply this logic to all providers

***To avoid a breaking change, the server must act exactly as it does today if this new env var is unset***.

**Declaration**:  

**Additional context**:  

This is a _great_ first issue for someone!  If you are new to Threat Dragon, reading the docs and understanding the config is a great first step!  You are also uniquely qualified to find blind-spots in the documentation as you fix this. 😎 

***This should not be done using Generative AI or an agent***.  "Good first issues" are meant for our human friends to learn and start participating.  If you have questions after being assigned the issue, I'm happy to help! Again, please read and understand the [contribution guide](../blob/main/contributing.md) before starting to work on this issue. You should be assigned before opening a pull request.

Below are some technical notes for new contributors to help get you started on this issue. I am sharing context and how _I would tackle it_ - not how it **needs** to be done. 😉

<details>
<summary>Provider OAuth Local Redirects</summary>

The provider-specific code lives in [td.server/src/providers](https://github.com/OWASP/threat-dragon/tree/main/td.server/src/providers).  Each provider has its own implementation.  In each implementation, we have the same/repeated redirect logic `getOauthReturnUrl`:

- [auth.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/controllers/auth.js#L25-L29)
- [github.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/providers/github.js#L49-L54)
- [google.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/providers/google.js#L31-L36)
- [gitlab.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/providers/gitlab.js#L38-L43)
- [bitbucket.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/providers/bitbucket.js#L46-L51)

I suggest creating a new oauth.js in the [helpers](https://github.com/OWASP/threat-dragon/tree/main/td.server/src/helpers) dir with a single function to replace all of the `getOauthReturnUrl` implementations. 

The function would first use `OAUTH_FRONTEND_RETURN_URL` if set, then fallback to the existing logic based on `NODE_ENV`.  I would remove the `getOauthReturnUrl` function from all providers and update any callers to use the new shared helper. 
</details>

<details>

<summary>Env Vars / Configuration</summary>

***The work for this issue belongs in the [env files](https://github.com/OWASP/threat-dragon/tree/main/td.server/src/env)***, not the configuration files.  Below is context and architecture information.

### Env Vars vs Config

Configuration is any runtime value that controls the available features and behavior of the server.  Environment variables inform the configuration. The [config dir](https://github.com/OWASP/threat-dragon/tree/main/td.server/src/config) includes static configs and the [Env loader](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/config/env.config.js).  We lean heavily on configuration to avoid breaking changes and maintain backwards compatibility.  Some configs might be viewed as feature flags - we do not make that distinction at this time.

### Env Architecture
There's a base [Env.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/env/Env.js) class that handles the loading logic: reading from the environment, loading a `.env` file, or reading values from a file (env var names can append `_FILE` to pass a file location instead of a raw value).

Because Threat Dragon supports unique feature sets per provider, we have split the Env files by high-level concern.  This is not limited to provider specific configuration. Each Env implementation extends the Env class.  This is meant to help with organization, standardize read logic, and reduce the amount of code required to add a new env var.

### Implementation Details

**You should _NOT_ create a new Env implementation for this work**.

I would add this new env var to [td.server/src/env/ThreatDragon.js](https://github.com/OWASP/threat-dragon/blob/main/td.server/src/env/ThreatDragon.js).  Because this config will impact all providers instead of a single one, it makes sense for it to live in the ThreatDragon config rather than repeating it for each provider.

Adding the new env var is as simple as adding a new object to the properties array. This new var should not be required or have a default. You can see how the env vars are imported and consumed by looking at the provider files that you'll be updating.

</details>

Adarshkumar0509:
## Summary:

Fixes #1539

I was looking at the providers and noticed all four of them had the exact same getOauthReturnUrl function copied. also NODE_ENV=development silently hard-codes the redirect to localhost:8080 which breaks if you deploy anywhere else.

## Description for the changelog:
Add OAUTH_FRONTEND_RETURN_URL config and shared OAuth return URL helper

## Other info:

changes:
- added OAUTH_FRONTEND_RETURN_URL to ThreatDragon.js (optional, no default)
- moved shared redirect logic into oauth.helper.js
- updated all 4 providers to use it
- added tests
- updated docs to explain what NODE_ENV does to oauth redirects

## Declaration:  

-Thanks for submitting a pull request, please make sure:  

- [X] content meets the [license](../blob/main/license.txt) for this project
- [X] appropriate unit tests have been created and/or modified
- [X] you have considered any changes required for the functional tests
- [X] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [X] *either* no AI-generated content has been used in this pull request

### DIFF
--- a/docs/configure/bitbucket.md
+++ b/docs/configure/bitbucket.md
@@ -34,7 +34,8 @@ and 'test' alters the functionality from what is being tested.
 
 - `NODE_ENV='development'`
 
-Note: when `NODE_ENV` is set to `development`, the OAuth redirect is hard-coded to `http://localhost:8080`. If the frontend runs on a different host or port, set `OAUTH_FRONTEND_RETURN_URL` to the correct base URL instead.
+Note: with `NODE_ENV=development`, the OAuth redirect points to `http://localhost:8080` by default.
+If your frontend runs elsewhere, set `OAUTH_FRONTEND_RETURN_URL` to the correct base URL.
 
 Server port number - this defaults to 3000, and it can be mapped to another port when running the docker command.
 So leave the server port at 3000 by not defining it, and it can then be mapped to external port 8080 using docker.
--- a/docs/configure/bitbucket.md
+++ b/docs/configure/bitbucket.md
@@ -34,6 +34,8 @@ and 'test' alters the functionality from what is being tested.
 
 - `NODE_ENV='development'`
 
+Note: when `NODE_ENV` is set to `development`, the OAuth redirect is hard-coded to `http://localhost:8080`. If the frontend runs on a different host or port, set `OAUTH_FRONTEND_RETURN_URL` to the correct base URL instead.
+
 Server port number - this defaults to 3000, and it can be mapped to another port when running the docker command.
 So leave the server port at 3000 by not defining it, and it can then be mapped to external port 8080 using docker.
 
--- a/docs/configure/configure.md
+++ b/docs/configure/configure.md
@@ -173,7 +173,7 @@ it depends on what repository access is required and whether running the desktop
 | `APP_TLS_CERT_PATH` | Locates the TLS certificate if TLS used | |
 | `APP_TLS_KEY_PATH` | Locates the TLS key if TLS used | |
 | `ENV_FILE` | The location of a dotenv file, if dotenv is used. Exported as it needs to be accessed before the dotenv file is read | `.env` |
-| `NODE_ENV` | Optional run mode: `production` enables rate limiting and relative OAuth redirects; `development` sets OAuth redirect to `http://localhost:8080`; `test` silences logging. Use `OAUTH_FRONTEND_RETURN_URL` to override the redirect instead of relying on this | `production` |
+| `NODE_ENV` | Optional run mode: `production`, `development` or `test` | `production` |
 | `LOCALES_ALLOWED` | Optional JSON array of allowed locale codes, e.g. `["en","es","fr"]`. When not set all locales will be available on the frontend | `[]` |
 | `LOCALE_DEFAULT` | Default locale code the server suggests, e.g. `pt`. When not set the server defaults to `en` | `en` |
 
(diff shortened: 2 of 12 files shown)


Adarshkumar0509:
done

jgadsden:
very good to see this elegant solution @Adarshkumar0509 
I will leave this for Leo to review as it is his domain
and I was not sure why the pull-request tests have not run for this PR?

Adarshkumar0509:
hii @jgadsden and @lreading actually this pull-request tests are running previously and passed all the test i do not know why suddenly it stopped i have seen some pull req and for them pr test was running well but for some of them pr test it was not running.

jgadsden (approved):
all changes look correct, and good to have this merged in
Many thanks @Adarshkumar0509

PR #1783, opened 2026-07-24
feat(server): add isObject and isValidThreatModel helpers to validators.helper.js

mahenoorsalat:
## Description
Adds \isObject\ and \isValidThreatModel\ validator helpers to \	d.server/src/helpers/validators.helper.js\ for threat model schema and payload validation.

## Features
- Adds \isObject\ type check helper.
- Adds \isValidThreatModel\ summary schema validation helper.

### DIFF
--- a/td.server/src/helpers/validators.helper.js
+++ b/td.server/src/helpers/validators.helper.js
@@ -1,3 +1,8 @@
 export const isString = (v) => typeof v === 'string';
 
 export const isNullish = (v) => v === null || v === undefined;
+
+export const isObject = (v) => typeof v === 'object' && v !== null && !Array.isArray(v);
+
+export const isValidThreatModel = (model) => isObject(model) && Boolean(model.summary && isString(model.summary.title));
+


jgadsden:
Hello @mahenoorsalat , could you be sure to follow our contributing guidelines before submitting a pull request?

lreading:
Could you explain the motivation for adding these helpers? They're not used, so it's just dead code as this PR sits, unless it's being used somewhere outside of Threat Dragon? Even at that, we have far more robust TM validation on the frontend.

`isObject` looks fine from a purely functional standpoint.
`isValidThreatModel` might be over-promising itself a bit though. Any object with `summary.title` would return true:

```javascript
const notAValidModel = {
  summary: { title: 'Definitely not a valid model' }
};

isValidThreatModel(notAValidModel); // true
```

lreading:
image

This is not a human contributor. 
You did not follow the contributing guidelines, use the PR template, identify your use of AI, and proposed a meaningless change that only added noise and work for the real humans who maintain this project.

If you open another PR like this, you will no longer be welcome to contribute.

Issue #1782, opened 2026-07-24
SBOM for frontend application missing in 2.6.2


rbdavison:
There is no SBOM for the frontend applications. i.e. I want to check the SBOM for the Windows version. However it is not available to download.

**Expected behaviour**:  

The sboms.zip file to be present in the Assets for download

**Environment**:  

The Github release page https://github.com/OWASP/threat-dragon/releases/#release-v2.6.2

**To Reproduce**:  

- Open the 2.6.2 release page: https://github.com/OWASP/threat-dragon/releases/#release-v2.6.2
- Expand the Assets section

Expected: The sboms.zip file to be present.
Actual: The sboms.zip file is not present.

**Declaration**:  

By submitting this issue you have:  

- [X] read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [X] not used agentic or generative AI in creating this bug report

**Any additional context, screenshots, etc**:  

<img width="961" height="721" alt="Image" src="https://github.com/user-attachments/assets/5ff09d46-8165-43e0-96a9-60fc8be27751" />

jgadsden:
does the file [sboms-container-image-app.zip](https://github.com/OWASP/threat-dragon/releases/download/v2.6.2/sboms-container-image-app.zip) provide this information?

rbdavison:
It does contain some SBOMs but they don't appear to be the versions for the standalone Windows / other applications.

<img width="638" height="164" alt="Image" src="https://github.com/user-attachments/assets/9d320c74-b0d4-4bb6-868e-2ffc22fe8c4d" />

The 2.6.1 sboms.zip contains the expected SBOMs.

<img width="730" height="349" alt="Image" src="https://github.com/user-attachments/assets/74726563-1f73-4a75-987a-7a1422825411" />

lreading:
I think I managed to mess this up during the 2.6.2 release.
Thanks for bringing it to our attention!

I'm going to try to create a one-off GH action that can generate SBOMs from the real release artifacts. I'd prefer that over me checking out the release and generating them locally or something - SBOMs should be auditable as much as possible. 😄

rbdavison:
100% agree with SBOMs being auditable.
it might also be good to put test cases in that the required SBOMs are present at the end of the process so if they fail to be generated in the future then the build will fail.
Thanks.

lreading:
[...]

PR #1793, opened 2026-07-27
Add GitHub action to regenerate SBOMs for a given release

lreading:
**Summary**:  

Relates to #1782 
The larger fix is probably taking a closer look at our release actions.

I think it's best to use an action to regenerate the sboms for a given release so that it's at least public and auditable.
This isn't a perfect solution because it's being done after-the-fact, but it should at least provide an accurate list of dependencies used in a given release.

chore: regenerate sboms for v2.6.2

**Declaration**:  

  - AI Tools: `Codex`
  - LLMs and versions: `I let it pick`
  - Prompts: `Investigate this issue, and build a one-off action that can reconstruct SBOMs for a given release`

**Other info**:  

Originally I was going to also make this a reusable workflow and try to fix the other release actions, but that scope felt a bit too big for what we need in the moment.

The AI also got really excited about our existing sbom zip not being a true zip but actually being a `tar` with a zip extension. I'll look into that eventually.

### DIFF
--- a/.github/workflows/sboms.yaml
+++ b/.github/workflows/sboms.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Validate requested release tag
         id: validate
         env:
-          RELEASE_TAG: ${{ inputs.release_tag || 'v2.6.2' }}
+          RELEASE_TAG: ${{ inputs.release_tag || 'v2.6.1' }}
         run: |
           if ! [[ "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z]+)*$ ]]; then
             echo 'release_tag must be a version tag such as v2.6.2' >&2
--- a/.github/workflows/sboms.yaml
+++ b/.github/workflows/sboms.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Validate requested release tag
         id: validate
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/sboms.yaml
```diff
@@ -0,0 +1,263 @@
...
+      - name: Check out requested release
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: refs/tags/${{ needs.validate_release_tag.outputs.release_tag }}
+          persist-credentials: false
+
+      - name: Resolve Node.js and npm versions
+        id: resolve
+        shell: bash
+        run: |
```
lreading, line 49: This is a bit messy, but using different versions of node/npm can change the sbom

lreading:
draft - pushing a test version that runs on push and has a hard-coded version tag in its env just to see if it works how i think it should. :)

lreading:
I'm pretty confident that this worked as intended.

Do we think there's value in keeping this action around with a workspace trigger, and/or use it for the future hardening of releases?

jgadsden (approved):
> Do we think there's value in keeping this action around with a workspace trigger, and/or use it for the future hardening of releases?

yes sure, worth keeping and it does no harm to keep it - you never know when we might need it again

PR #1862, opened 2026-09-08
Add i18n lint

lreading:
**Summary**:  

Closes #1859 
The linter is configured to warn on dynamic translations. I will be changing that to error in a future PR that also removes the dynamic translations.
This should help prevent future translation mishaps (not that I've ever caused any....)! :smile: 

**Description for the changelog**:  

chore: add translation linting
bugfix: fix missing translations

**Declaration**:  

Thanks for submitting a pull request, please make sure:  

- [x] content meets the [license](../blob/main/license.txt) for this project
- [x] appropriate unit tests have been created and/or modified
- [x] you have considered any changes required for the functional tests
- [x] you have read the [contribution guide](../blob/main/contributing.md) and agree to the [Code of Conduct](../blob/main/code_of_conduct.md)
- [ ] *either* no AI-generated content has been used in this pull request
- [x] *or* any [use of AI](../blob/main/contributing.md#use-of-ai) in this pull request has been disclosed below:
  - AI Tools: `ChatGPT`
  - LLMs and versions: `Unsure, it's kinda opaque now`
  - Prompts: `For Threat Dragon, is there a way to lint for missing translation strings?`

**Other info**:

PR #1863, opened 2026-09-08
Remove dynamic translations

lreading:
**Summary**:  

***This PR looks bigger than it is because I based it on #1862***
I did this as 2 separate PRs to keep them manageable. After #1862 is merged, I will rebase and the size of the PR will go down quite a bit. :) 

Removes dynamic translation strings so that we can do a full i18n lint
Updates linting rules to error on dynamic translation strings
This should prevent us from accidentally missing translation strings in the future

chore: remove dynamic translation strings

**Declaration**:  

  - AI Tools: `Codex`
  - LLMs and versions: `gpt-5.6-terra`
  - Prompts: `Update the eslint config to error on dynamic translation strings. Find and fix all errors. Use a shared helper to do translation string lookups`

**Other info**:  

Depends on #1862  - was in draft until that was merged, now ready for review


td.vue/src/service/i18n/translation.js
```diff
@@ -0,0 +1,114 @@
+const translateKnownKey = (t, key) => {
+    switch (key) {
+    case 'branch.protectedBranch': return t('branch.protectedBranch');
```
lreading, line 3: The linter checks for the literal `$t(..)` syntax (along with others). 
If you had a simple switch with just strings, it would not know those were translations and they would not be linted.

So this looks a bit ugly/awkward, but I promise it's for a good reason! 😆

jgadsden: true, it looks awkward but I agree it is there for good reason

jgadsden (approved):
visually tested locally (in both English and Greek) and all strings test OK, with no untranslated strings

Issue #1393, opened 2025-11-25
Optimize HTML/PDF report for viewing on mobile devices


mingchen:
**Describe what problem your feature request solves**:
The generated HTML/PDF report isn’t optimized for viewing on mobile devices, such as phones with small screens.

**Describe the solution you'd like**:
Generate a mobile-optimized page layout that adapts to different screen sizes.

**Additional context**:
To preview how the current report appears on a small screen, simply narrow your browser window—you should immediately see the effect.

jgadsden:
Hello @mingchen , thanks for the enhacement request. Do you have a possibe implementation? All pull requests are appreciated

jgadsden:
yes, sure - and take into account the restrictions on the [use of (generative) AI](https://github.com/OWASP/threat-dragon/blob/main/contributing.md#use-of-ai) that all our contributors to issues + comments + pull requests are required to follow

lreading:
I'm going to close this. While it'd be nice to have, the vast majority of users are not using Threat Dragon on mobile. I could kind of see someone _maybe_ viewing a report on mobile, but I seriously doubt that anyone is contributing to a threat model on mobile.

jgadsden:
yes, agreed, it is not something that is requested by the community

PR #1880, opened 2026-09-17
Fix mobile layout overflow on home and selection pages

christianbaumann:
On phones the home page text and logo hung outside the grey hero box, page scrolled sideways.

Two causes:

- `.td-description` had a hardcoded `margin-left: 170px` to clear the logo on desktop. Applied at every breakpoint, so once the `md` columns stack it shoves the paragraph out of the box.
- `TdImage` had no `max-width`, and the logo SVG is 400px wide.

Fix: indent only from `768px` up, `TdImage` gets `max-width: 100%`, dropped the logo's fixed side margins (column gutter already does that job).

Same mistake elsewhere: unprefixed `offset` / `ml-5` also apply at all breakpoints. Fixed in `SelectionPage.vue`, `ImportModel.vue`, `SelectDemoModel.vue`.

7 unit tests, 6 Cypress mobile tests. All checked red before the fix.

Reviewers: `home.png` visual baseline needs updating, logo no longer overflows its column on desktop either.


jgadsden:
@christianbaumann you need to read the contributors guidelines first, in particular the section on AI, and the code of conduct
your pull request is clearly AI generated
we have recently closed #1393 which discusses this subject, please read this issue first

